Issue #11, opened 2025-07-14
Secret lifetimes


domenkozar:
Ideally, secrets live for a certain amount of time, then it's time to request a new secret.

There are two aspects where this becomes tricky:

- keep the state of how long the secret is still valid for
- provide an api when to reload the app after a new secret is fetched

The API would look something like:

```rust
let secrets = SecretSpec::builder().load()?;

for secret in secrets.next() {
   ...
}

for mysecret in secrets.mysecret.next() {
   ...
}
```

blaggacao:
> provide an api when to reload the app after a new secret is fetched

Or send a process signal, which may be understood by the app as instructions to just reload the credential context, instead the entire app.

blaggacao:
> keep the state of how long the secret is still valid for

Or fetch validity information from a revocation mechanism - this may be an advanced use case and not necessary for many scenarios, but it is critical enough in others that it might be worth considering plugging into some sort of established standard.

domenkozar:
Pulling threads from the discussion together. I think this naturally splits into three shippable slices:

**1. Provider TTL tracking.** Extend the `Provider` trait with something like:

```rust
fn get_with_lease(&self, key: &str) -> Result<(String, Option<Lease>)>;
```

Static providers (keyring, dotenv, 1Password) return `None`. Lease-capable providers (Vault dynamic secrets, AWS STS, GCP IAM tokens) return a real expiry plus optional renew handle. Everything else builds on this.

**2. SDK refresh API.** Once lease info exists, the SDK exposes a stream of updates. The iterator shape from the issue body works, though I'd lean async so apps only re-wire the specific key that rotated:

```rust
let mut secrets = SecretSpec::builder().load().await?;

while let Some(update) = secrets.next_refresh().await {
    reload_db_pool(&update.database_url);
}
```

Static secrets simply never emit, so the same code is safe for any provider mix.

**3. Reload contract for `secretspec run`.** For wrapped commands the CLI owns the signal. SIGHUP on rotation (per @blaggacao) is the right default since it matches the nginx / postgres / haproxy convention. An opt-in `--reload-signal SIGUSR1` flag covers apps that bind HUP elsewhere. Re-exec of the child works as a fallback for apps that can't handle signals at all, behind a flag.

Prior art worth borrowing from:
* Vault Agent: clean split between fetch lifecycle and app signaling, maps almost 1:1 to (1)+(2) vs (3).
[...]

PR #19, opened 2025-07-22
feat: integrate secrecy crate for secure secret handling

domenkozar:
- Add secrecy v0.10.3 dependency with serde feature
- Update Provider trait to use SecretString for get/set operations
- Update all provider implementations to handle SecretString
- Update ValidatedSecrets to store secrets as SecretString
- Modify get, set, and run commands to safely handle secrets
- Update derive macro to convert between SecretString and String

This ensures secrets are protected in memory with automatic zeroing when dropped, preventing accidental exposure through debug prints or logs. The generated structs from the derive macro continue to use regular String types for backward compatibility and serialization support.

🤖 Generated with [Claude Code](https://claude.ai/code)


secretspec-derive/src/lib.rs
```diff
@@ -524,13 +524,14 @@ fn generate_secret_assignment(
 ) -> proc_macro2::TokenStream {
     if is_optional {
         quote! {
-            #field_name: #source.get(#secret_name).cloned()
+            #field_name: #source.get(#secret_name).map(|s| s.expose_secret().to_string())
```
domenkozar, line 527: We should probably keep the Secrecy wrapper.

domenkozar: Let's leave strings for now, I think we should design our own wrapper that's going to offer a bunch of other things like #11

Issue #29, opened 2025-08-04
SDK: Please make it possible to use typed secret names (macro-generated, not strings), without loading all secrets at startup


emchristiansen:
I'm using this pattern to load SS in Rust (note there's a macro bug workaround in there, too):

```rust
use std::sync::OnceLock;

use secretspec::Resolved;

// Generate typed structs from secretspec.toml
secretspec_derive::declare_secrets!("../secretspec.toml");

// Constants for default provider and profile
pub const PROVIDER: &str = "onepassword://SecretSpecMetacortex";
pub const PROFILE: &str = "default";

// Global SecretSpec instance
static SECRETSPEC: OnceLock<Resolved<SecretSpec>> = OnceLock::new();

/// Get a reference to the global SecretSpec instance
/// Lazily initializes on first access
///
/// NOTE: This uses a workaround for a bug in secretspec where the builder pattern
/// strips vault information from onepassword URIs. The builder extracts only the
/// provider name (e.g., "onepassword") from URIs like "onepassword://VaultName",
/// losing the vault specification.
///
/// TODO: Switch to builder pattern when the bug is fixed:
/// ```
/// SecretSpec::builder()
///   .with_provider(PROVIDER)
///   .with_profile(Profile::Default)
///   .load()
/// ```
///
/// Current workaround uses side-effect based API where `secretspec::Secrets`
/// is a global singleton that holds configuration. When we call set_provider()
/// and set_profile(), we're modifying global state that SecretSpec::load()
/// then reads from.
///
/// PERFORMANCE WARNING: Initial load is VERY slow (~40s for 39 secrets) because
/// SecretSpec appears to make individual `op` CLI calls for each secret instead
/// of batching them. The OnceLock ensures this only happens once per process.
pub fn secretspec() -> &'static Resolved<SecretSpec>
{
  SECRETSPEC.get_or_init(|| {
    // Load mutable Secrets and set provider/profile directly to preserve vault info
    let mut spec =
      secretspec::Secrets::load().expect("Failed to load Secrets");
    spec.set_provider(PROVIDER);
    spec.set_profile(PROFILE);

    // Now load SecretSpec using the configured spec
    // This will use the provider/profile we just set
    SecretSpec::load(None::<String>, None::<Profile>)
      .expect("Failed to load SecretSpec")
  })
}
```

I want to use `SecretSpec` for the type safety (`Secret` expects string args), but it loads all my secrets at startup.
With OnePassword this is very slow!

[...]

PR #30, opened 2025-08-05
fix: preserve provider URI details in builder pattern

emchristiansen:
Related to: https://github.com/cachix/secretspec/issues/29

fix: preserve provider URI details in builder pattern

Previously, the builder pattern would convert provider URIs to Provider
instances and then only extract the provider name (e.g., "onepassword")
instead of preserving the full URI with vault/path information
(e.g., "onepassword://SecretSpecMetacortex").

This fix:
- Adds a uri() method to the Provider trait that returns the full URI
- Implements uri() for all providers following the SecretSpec URI spec:
  - OnePassword: "onepassword://[account@]vault" or "onepassword+token://vault"
  - Dotenv: "dotenv" or "dotenv:path" (single colon format)
  - Keyring: "keyring"
  - Env: "env"
  - LastPass: "lastpass" or "lastpass://folder"
- Updates the builder to use provider.uri() instead of provider.name()
- Updates Resolved struct to store the full URI instead of just the name

This ensures that vault names, paths, and other provider-specific
configuration are preserved when using the builder pattern, fixing
issues where secrets couldn't be found due to missing vault information.

### DIFF
--- a/secretspec-derive/src/lib.rs
+++ b/secretspec-derive/src/lib.rs
@@ -1199,8 +1199,8 @@ mod builder_generation {
             let provider_str = if let Some(provider_fn) = #provider_expr {
                 let provider_box = provider_fn()
                     .map_err(|e| secretspec::SecretSpecError::ProviderOperationFailed(e))?;
-                // Get the provider name to pass as a string to set_provider
-                Some(provider_box.name().to_string())
+                // Get the full URI to pass as a string to set_provider (preserves vault info)
+                Some(provider_box.uri())
             } else {
                 None
             };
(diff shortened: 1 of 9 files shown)


emchristiansen:
This code was written with Claude Code.

domenkozar:
We should be using `Into<String>` and `Into<Uri>` here instead of methods.

Also, need to come with some tests per provider, Claude should be able to write those up.

domenkozar:
@emchristiansen ping

emchristiansen:
Hey sorry, I'm not sure if I'll have time to push this through!
Feel free to take over if you wish!

PR #75, opened 2026-04-19
fix: fall back to default profile when requested profile is undefined

ap-1:
When a profile is requested that isn't explicitly defined in `secretspec.toml`, `resolve_profile` errors with "Profile 'X' not found" instead of falling back to `[profiles.default]`.

This means projects that only need a single set of secrets across all environments are forced to declare empty profile sections for every profile name that might be used:

```toml
[profiles.default]
API_KEY = { required = true }

# to avoid "not found" errors
[profiles.dev]
[profiles.staging]
[profiles.prod]
```

This PR changes `resolve_profile` to fall back to the `default` profile when the requested profile doesn't exist, matching the intuition that `default` applies universally, not just when no profile is specified.

Provider reads also fall back to the default profile. When a secret is not found at the requested profile's storage path (e.g., `secretspec/myapp/prod/API_KEY`), the provider retries at the default path (`secretspec/myapp/default/API_KEY`). This ensures secrets stored under default are accessible from any profile without duplication.

The error is preserved for the case where neither the requested profile nor `default` exists. The error message is also improved to list all available profiles, making it easier to fix any misconfiguration.

### DIFF
--- a/secretspec/src/secrets.rs
+++ b/secretspec/src/secrets.rs
@@ -228,9 +228,19 @@ impl Secrets {
             .config
             .profiles
             .get(&profile_name)
+            .or_else(|| self.config.profiles.get("default"))
             .cloned()
             .ok_or_else(|| {
-                SecretSpecError::SecretNotFound(format!("Profile '{}' not found", profile_name))
+                SecretSpecError::SecretNotFound(format!(
+                    "Profile '{}' is not defined in secretspec.toml. Available profiles: {}",
+                    profile_name,
+                    self.config
+                        .profiles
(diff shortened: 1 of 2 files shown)


domenkozar:
I'm not entirely sure I want to make this implicit, mostly because what if it's a mistake like a typo?

Is there a good reason you can't put empty profile into the spec?

ap-1:
Just QoL, I wanted to avoid having to write them out manually if I could

domenkozar:
let's not do this, I prefer to be explicit.

Issue #64, opened 2026-03-10
Support out-of-tree providers via gRPC interface


guilhem:
Propose a standardized interface (e.g., gRPC) to allow implementing out-of-tree (external) providers in secretspec.

Use case: secret stores that are internal, private or closed source, enabled via an independent plugin (not part of upstream/open source).

Example scenarios:
- secretspec contacts an external provider via a gRPC or similar interface
- This enables organizations to add secret providers without exposing code
- Pattern similar to secret management integrations like Vault, 1Password, SOPS, etc.

This would make it easier to integrate proprietary or internal secrets sources without patching secretspec directly.

Thanks!

schickling-assistant:
+1 — landed here while working around exactly this gap. Concrete use case + a sketch of an alternative transport that might be lower-friction than gRPC for the most common case.

## Our setup

We have an internal tool `op-proxy` that wraps the 1Password CLI with approval gates, audit logging, and a local cache. It's intentionally machine-local — every read is `op-proxy read --reason "<text>" --cache <ttl> op://Vault/.../field`, and the reason gets surfaced to a desktop approval prompt or recorded in an audit log.

`SECRETSPEC_OPCLI_PATH` doesn't fit:
- `op-proxy`'s CLI shape differs from `op` (`op-proxy read -r <reason> <op://...>` vs `op item get --format=json <name>`).
- The built-in `onepassword://` provider owns its own naming (`secretspec/{project}/{profile}/{key}`) — we read arbitrary `op://Vault/...` refs.
- `--reason` is required on every call; secretspec can't supply one through the current trait.

So we patched a real `opproxy://` provider into a fork (~250 lines of Rust modeled on `env.rs`/`onepassword.rs`, registered through the existing `linkme::distributed_slice`). Works great as a fork, but we don't want to maintain a Nix overlay rebase against every upstream release, and the provider has no place in upstream because op-proxy is intentionally machine-local.

## On the transport

gRPC works but introduces a real dependency footprint (tonic, prost, schema versioning, codegen) and a long-lived subprocess lifecycle problem (when does the plugin server start/stop? how does it authenticate to the host? how does it pass `--reason`-style per-call context cleanly?).

Two cheaper alternatives worth considering:

[...]

PR #98, opened 2026-05-28
docs: draft Secret Provider Protocol v1

domenkozar:
## Summary
- Drafts a `reference/provider-protocol` docs page proposing the wire format for out-of-tree providers (#64)
- Plugins are discovered as `secretspec-provider-<scheme>` binaries on `$PATH` and exchange line-delimited JSON over stdio
- Sidebar entry labelled "Provider Protocol (draft)" so readers know it isn't shipped yet

The page itself carries a draft warning at the top: the protocol is specified, not yet implemented in the host. Goal of this PR is to lock the wire contract on paper so a reference plugin and host integration can land against a stable target.

## Test plan
- [ ] `cd docs && npm install && npm run dev`, verify the new page renders and the sidebar entry shows under Reference
- [ ] Confirm anchors in section links (#11, #64) point to the right issues
- [ ] Review against @schickling-assistant's `opproxy://` use case (per-call `context.reason`, capability-gated `set`/`reflect`) on #64

Closes: nothing yet. Once merged the next steps are (a) reference Rust plugin under `examples/provider-plugin/`, (b) host-side plugin loader keyed off URI scheme.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -155,6 +155,10 @@ Secrets can be stored in: keyring (default), dotenv files, environment variables
             { label: "CLI Commands", slug: "reference/cli" },
             { label: "Providers", slug: "reference/providers" },
             { label: "Adding Providers", slug: "reference/adding-providers" },
+            {
+              label: "Provider Protocol (draft)",
+              slug: "reference/provider-protocol",
+            },
           ],
         },
       ],
--- a/docs/src/content/docs/reference/provider-protocol.md
+++ b/docs/src/content/docs/reference/provider-protocol.md
@@ -1,0 +1,280 @@
+---
+title: Secret Provider Protocol (v1)
+description: Wire-format specification for out-of-tree secretspec providers, exchanged as line-delimited JSON over stdio.
+---
+
+A wire-format specification for out-of-tree secretspec providers. Plugins implementing this protocol are discovered as subprocess binaries on `$PATH` and exchange line-delimited JSON over stdio. The protocol is transport-agnostic at the JSON level; the v1 transport is stdio.
+
+This document uses the keywords MUST, SHOULD, and MAY per RFC 2119.
+
+:::caution
+This specification is a **draft proposal**. It is not yet implemented in the host. See [issue #64](https://github.com/cachix/secretspec/issues/64) for the design discussion.
+:::
+
+## 1. Discovery
+
+A provider URI scheme `<scheme>://...` is bound to a plugin binary named `secretspec-provider-<scheme>`. The host locates the binary by searching `$PATH` in order; the first match wins.
+
+The scheme MUST match `^[a-z][a-z0-9_-]*$`. Hyphens in the scheme map literally to hyphens in the binary name (`opproxy://` resolves to `secretspec-provider-opproxy`).
+
+If no binary is found, the host MUST return an error to the caller distinguishing "plugin not installed" from "plugin returned an error."
+
+## 2. Invocation model
+
+The host spawns the plugin once per session. A session corresponds to a single resolution pass (e.g., one `secretspec run`, one `secretspec check`). Within a session the plugin MAY process any number of requests.
+
+On spawn, the host sets the following environment variables:
+
+| Variable | Value |
+|---|---|
+| `SECRETSPEC_PROTOCOL_VERSION` | Highest protocol version the host supports (e.g., `1`) |
+| `SECRETSPEC_PROVIDER_URI` | The full URI that selected this plugin |
+
+No command-line arguments are passed in v1. The host MUST NOT pass secrets via `argv` or environment.
+
+The plugin reads newline-delimited JSON requests from stdin and writes newline-delimited JSON responses to stdout. The plugin exits on stdin EOF or after responding to a `bye` request. Stderr is free-form and the host MUST NOT parse it; plugins SHOULD use stderr for diagnostic output only and MUST NOT write secret values to stderr.
+
+The host closes stdin to signal end-of-session. The plugin SHOULD exit within 5 seconds of stdin EOF; the host MAY send `SIGTERM` and then `SIGKILL` after a grace period.
+
+## 3. Wire format
+
+Each request and each response is exactly one JSON object terminated by a single `\n`. Embedded newlines inside JSON strings MUST be escaped per RFC 8259. The encoding is UTF-8.
+
+The host MUST NOT pipeline: it sends one request, waits for the matching response, then sends the next. Responses MUST appear on stdout in the same order as their requests.
+
+## 4. Handshake
+
+The first request in every session is `hello`. The plugin MUST NOT process any other request before responding to `hello`.
+
+**Request:**
+
+```json
+{
+  "op": "hello",
+  "protocol_version": 1,
+  "uri": "opproxy://vault/Production?reason=build",
+  "context": { "reason": "building api image" }
+}
+```
+
+**Successful response:**
+
+```json
+{
+  "ok": true,
+  "protocol_version": 1,
+  "name": "opproxy",
+  "capabilities": ["get", "set", "batch_get", "reflect"]
+}
+```
+
+The plugin's `protocol_version` MUST be less than or equal to the host's. If the plugin cannot support the host's version, it responds with an error of kind `unsupported_version` and exits.
+
+The plugin advertises its supported operations in `capabilities`. The host MUST NOT send an operation absent from this list. `get` is mandatory; all others are optional.
+
+## 5. Operations
+
+All requests carry `"op": "<name>"`. All successful responses carry `"ok": true`; failures carry `"ok": false` and an `error` object (see section 6).
+
+### 5.1 get
+
+Retrieves a single secret.
+
+**Request:**
+
+```json
+{ "op": "get", "project": "myapp", "key": "DATABASE_URL", "profile": "production" }
+```
+
+**Response (hit):**
+
+```json
+{ "ok": true, "value": "postgres://..." }
+```
+
+**Response (miss):**
+
+```json
+{ "ok": true, "value": null }
+```
+
+A miss is distinct from an error. Missing secrets MUST return `value: null`, not an error.
+
+### 5.2 set
+
+Stores a single secret. Only callable if the plugin advertised the `set` capability.
+
+**Request:**
+
+```json
+{ "op": "set", "project": "myapp", "key": "DATABASE_URL", "value": "postgres://...", "profile": "production" }
+```
+
+**Response:**
+
+```json
+{ "ok": true }
+```
+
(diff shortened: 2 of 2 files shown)


schickling-assistant:
Thanks for drafting this. The direction matches the `opproxy://` use case much better than a gRPC-first design: PATH-discovered stdio helpers are exactly the shape that would let us delete our patched `secretspec` overlay and ship `op-proxy` as an out-of-tree provider.

From our current fork, these are the requirements I would like the draft to make explicit before it hardens:

1. Pass the loaded SecretSpec file path to the plugin.

   Our provider does not derive the backing external-secret refs from the URI alone. It reads the repo's committed `secretspec.toml`, specifically an extension table:

   ```toml
   [x-op-proxy.refs]
   APP_TOKEN = "external-secret-ref-for-app-token"
   BOT_TOKEN = { ref = "external-secret-ref-for-bot-token", reason = "bot-runtime", cache = "12h" }
   ```

   Today our patched provider has to rediscover that file via URI path -> `SECRETSPEC_FILE` -> cwd walk because the in-process provider trait does not tell providers which file the host loaded. For out-of-tree providers, please make this first-class, e.g. include one of these in `hello`:

   ```json
   {
     "op": "hello",
     "protocol_version": 1,
     "uri": "opproxy://",
     "config_file": "/abs/path/to/secretspec.toml",
     "context": { "reason": "run local app" }
   }
   ```

   or define `SECRETSPEC_FILE=/abs/path/to/secretspec.toml` as a required env var for plugin sessions. I prefer the explicit `hello.config_file` field because it avoids hidden env coupling, but either would work.

2. Pass extension metadata, or explicitly bless plugin-owned extension tables.

   `opproxy` needs per-key backend refs and per-key overrides (`reason`, `cache`). It would be useful for the host to either:

   - send unknown/extension metadata for the requested key in `get` / `batch_get`, or
   - document that plugins MAY read plugin-owned tables such as `[x-op-proxy.refs]` from `config_file`.

   If the latter is the intended model, the protocol should say unknown `x-*` tables are reserved for tool/provider extensions and are safe for plugins to inspect.

3. Add provider URI or config to every request, or state it is session-constant.

   The current draft puts `uri` in `hello` only. That is fine if one spawned plugin session is strictly bound to one provider URI, but the spec should say that explicitly. An `opproxy://?cache=12h` session may parse defaults from the URI once and apply them to all later `get` calls.

4. Clarify per-command context versus provider defaults.

   `context.reason` is the load-bearing bit for approval/audit, so this is good. Two small improvements would help:

   - Define whether `context` can also be supplied to `check` and `get`, not just `run`. `opproxy` needs a reason for any backend read, including `secretspec check` / `batch_get`.
   - Recommend host-generated fallback context for non-interactive commands, e.g. `reason = "secretspec:<project>:<key>"` or `"secretspec:<project>:check"`, while still allowing plugins to reject missing context. Without this, `check` may become awkward for providers that require audit reasons.

5. Keep `batch_get`, and allow plugins to report per-key misses cleanly.

   This is important for approval-gated backends because it can reduce repeated prompts and allow provider-side caching. The current shape is good. Please keep the rule that missing keys are `null`, not errors.

6. Add host behavior for secret redaction around plugin failures.

   The security section already says plugins must not write secrets to stderr. The host side should also say plugin stderr and structured error messages may be surfaced to users, so plugins MUST treat both as non-secret, and the host SHOULD avoid including request values in debug logs. This matters more once `set` exists.

If those are covered, the protocol should absorb the rest of our current fork. The migration path would be straightforward: replace the patched Rust provider with a `secretspec-provider-opproxy` binary that reads `config_file`, resolves `[x-op-proxy.refs]`, uses `context.reason` plus per-secret overrides, and calls `op-proxy read` with the requested reason/cache policy.

One caveat outside this PR: this only removes our overlay once the host-side plugin loader lands and downstream consumers such as `devenv` link a `secretspec` version containing it. Today `devenv` links `secretspec` as a Rust library, so our `pkgs.secretspec` overlay cannot affect `secretspec.enable = true`; that is why we currently keep native devenv SecretSpec disabled and use the patched standalone binary path.


<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_name` | 🦈 co2-lagoon |
| `agent_session_id` | bebc51af-803e-4e7f-91f3-60a9f39cfbfa |
| `agent_tool` | Codex CLI |
| `agent_tool_version` | 0.130.0 |
| `agent_runtime` | Codex CLI 0.130.0 |
| `agent_model` | unknown |
| `worktree` | dotfiles/main |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@unknown-dirty |
</details>

domenkozar:
Thanks @schickling-assistant — this is exactly the kind of grounding the draft needed. Pushed a revision (a2545b1) that folds in all six points:

1. **Config file path → plugin.** `hello` now carries an explicit `config_file` (absolute path, or `null` for a purely programmatic SDK caller), mirrored as `SECRETSPEC_FILE` for plugins that prefer env. Went with the explicit field as the primary, per your preference for avoiding hidden env coupling. No more URI → `SECRETSPEC_FILE` → cwd-walk rediscovery.

2. **Extension metadata.** New §4.1 blesses the plugin-owned model: `x-*` tables are reserved for tool/provider extensions, the host MUST NOT interpret them and MUST preserve them on rewrite, and plugins MAY read them (e.g. `[x-op-proxy.refs]`) from `config_file`. I chose "plugins read their own table" over "host forwards metadata per-key" for v1 — simpler host, and the plugin owns/validates its own shape. Open to the per-request forwarding variant in v2 if it proves limiting.

3. **URI session-constancy.** §2 now states a session is bound to exactly one provider URI for its lifetime, the host MUST NOT reuse a process across URIs, and the plugin MAY parse the URI (incl. `?cache=12h`) once at `hello`. The URI is not repeated per request.

4. **Per-command context.** §7 now says the session context applies to *every* op in the pass — `check`, `get`, `batch_get`, `import`, not just `run`. It also recommends the host synthesize a fallback `context.reason` (`secretspec:<project>:check`, `secretspec:<project>:<key>`, etc.) when the caller gives none, with caller-supplied reason taking precedence and plugins still free to reject.

5. **batch_get / null misses.** Kept as-is — §5.3 already mandates missing keys appear in `values` as `null`, never as errors. No change needed, just confirming it stays.

6. **Redaction around failures.** §8 now says the host MAY surface plugin stderr and `error.message` to users/logs, so plugins MUST treat both as non-secret and MUST NOT embed values (called out specifically for `set`), and the host SHOULD keep request/response values out of debug logs.

On the `devenv`-links-secretspec-as-a-library caveat: agreed that's orthogonal to this PR and only resolves once the host-side loader lands and downstream bumps the version. This PR is still just the paper spec; the loader + `examples/provider-plugin/` reference are the follow-ups.

ap-1:
@domenkozar I mentioned this briefly in the Discord, but given you're working on a complete spec here I'll provide a more complete description copied from a conversation I had with @claude:

**Use case: write-only CI secret sinks (GitHub Actions / Forgejo repo secrets).**

I'd like to declaratively push a repo's CI secrets from a source of truth (OpenBao) using the same `secretspec.toml` that already declares them. The PATH-discovered stdio shape fits this well — a `secretspec-provider-github` adapter over the repo secrets API. But these backends are a provider *class* the v1 draft can't currently express: **write-only sinks**. The GitHub/Forgejo Actions API lets you create/update a secret, delete it, and list secret *names* — it never returns a value. No read-back, by design.

Three things I'd ask the draft to make explicit before it hardens:

**1. `get` must be optional — allow a capability set without it.**

§4 states "`get` is mandatory; all others are optional." A write-only sink can never implement `get`. Forcing it means the plugin returns `value: null` for every key, which is indistinguishable from "secret genuinely absent" — so `secretspec run`/`get` would silently inject nothing while reporting everything missing. That's the exact silent-data footgun a capability model should prevent.

Proposed: drop the "`get` is mandatory" rule; require only that a plugin advertise ≥1 operation. A write-only sink advertises:

```json
{ "ok": true, "protocol_version": 1, "name": "github", "capabilities": ["set", "exists", "delete", "reflect"] }
```

and the host MUST refuse **value-read** operations (`run`, `get`, and the SDK's value access) against a provider whose capabilities omit `get`, failing with a clear "this provider is write-only" error rather than treating its secrets as `null`/absent. (This is the wire analog of the `allows_get()` trait method you said was in-scope for in-tree providers; keeping the two models symmetric matters, since the same host commands drive both.)

**2. Presence-without-value: add an `exists` op, alongside `reflect`.**

A write-only sink can answer "is this key set?" (list names) even though it can't answer "what is its value?". `import`'s "already exists in target" check, `check`'s presence verification, and any future reconcile all need that presence signal *without* reading a value. Two operations cover it:

- A capability-gated `exists` op for a targeted single-key check:

```json
{ "op": "exists", "project": "myapp", "key": "DATABASE_URL", "profile": "production" }
```

```json
{ "ok": true, "exists": true }
```

- §5.4 `reflect` for enumeration (it already returns the name set for a project).

Please state that the host's presence checks use `exists`/`reflect`, never `get`, so:

- `secretspec import` into a write-only destination skips already-present keys (per-key `exists`, or one `reflect` + membership), and
- `secretspec check` against a write-only sink verifies the declared secrets are actually in the repo, instead of being refused.

For a write-only sink both `exists` and `reflect` map to the same list-names API call; `exists` is the cheap targeted form, `reflect` the bulk form. Gate `exists` behind its own capability like `set`/`reflect`. (`exists` returns a boolean, so there's no `null`-vs-error ambiguity to resolve — it's distinct from a `get` miss.) `exists` is also the wire analog of the `exists()` trait method discussed for in-tree providers, keeping the two models aligned.

**3. Add a capability-gated `delete` operation.**

There's no `delete` in v1, and it isn't in §10's non-goals either. Declarative management means pruning a secret from the repo when it's removed from `secretspec.toml` — and for a write-only sink a stale secret is *active state* (it keeps getting injected into CI runs) that you can't even see by reading. That's the one case where leftover backend state actually matters. Reserve the op in the wire format now, even if the host-side `sync`/reconcile command is a follow-up, so adding it later doesn't force a version bump:

```json
{ "op": "delete", "project": "myapp", "key": "OLD_TOKEN", "profile": "production" }
```

```json
{ "ok": true }
```

Gate it behind a `delete` capability like `set`; deleting an absent key SHOULD be an idempotent no-op success.

---

With (1)–(3), a `secretspec-provider-github` plugin advertising `["set", "exists", "delete", "reflect"]` lets `secretspec import openbao://…` push declared secrets into a repo and a later `secretspec sync` prune removed ones — all driven by the committed `secretspec.toml`, with no read-back ever attempted. The rest of the draft (session model, `context.reason`, the `set`-era security notes in §8) already fits this use case unchanged.

domenkozar:
Going to come back to this.

domenkozar:
See https://github.com/cachix/secretspec/pull/362

PR #121, opened 2026-07-06
feat: native secret references (`ref`) as provider independent coordinates

### LINKED ISSUE #64: Support out-of-tree providers via gRPC interface

domenkozar:
Closes #64.

A secret can name one externally managed secret by its store's own coordinates, replacing SecretSpec's `{project}/{profile}/{key}` naming for that secret:

```toml
[profiles.production]
DATABASE_URL = { description = "Postgres DSN", ref = { item = "db", field = "password" }, providers = ["prod_op"] }
INFRA_TOKEN  = { description = "Infra token", ref = { vault = "Production", item = "infra", field = "token" } }
GITHUB_TOKEN = { description = "GitHub token", ref = { item = "GITHUB_PAT" }, providers = ["env"] }
```

`item` is required; `field` (1Password field label, Vault KV field, AWS JSON key, keyring account), `vault`/`section` (1Password), and `version` (GCSM) refine it where the store supports them. All 11 providers resolve refs and reject coordinates they have no equivalent for.

## Design: refs name, providers route

This PR went through a design pivot mid-review. The first iteration carried references as provider URIs (`ref = "op://vault/item/field"` plus per provider grammars like `?item=`, `?entry=`, `?field=`). That fused two independent axes, naming and routing, into one string: single secrets impersonated whole provider instances, precedence needed a special "ref beats --provider" rule, and `ref` had to be mutually exclusive with `providers`.

The final design makes `ref` a structured table of pure naming coordinates:

- **Routing is unchanged and uniform.** Which store resolves the coordinates follows the same rules as every secret: `--provider`/`SECRETSPEC_PROVIDER` override, then the secret's `providers` chain (entries may now be inline URIs, no alias needed), then defaults. So refs compose with fallback chains, and an override redirects them, e.g. `--provider dotenv:.env.fixtures` during tests.
- **Writes are symmetric.** `secretspec set` and `check` prompting write through the coordinates in place where the backend allows it (1Password, keyring, pass, dotenv, bws, Proton Pass, LastPass); Vault/AWS/GCSM refs are read only.
- **Pasted URIs translate themselves.** `ref = "op://Production/db/password"` fails with the exact table to write instead. Provider URIs stay store addresses only.
- **Internals:** the `Provider` trait now takes an `Address` (convention naming or native coordinates), which deleted the instance level reference machinery: `supports_references()`, 11 reference URI grammars, fail closed write guards, and batch fan out patches. Secrets sharing coordinates fetch once; audit events record the coordinates in a new `ref` field; manifest validation now runs on every load.

The [configuration reference](docs/src/content/docs/reference/configuration.md) documents the coordinate table and the per provider translation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,76 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`ref`: native secret references on secrets**: a secret can name one
+  externally managed secret by its store's own coordinates, instead of
+  SecretSpec's `{project}/{profile}/{key}` naming:
+
+  ```toml
+  [profiles.production]
+  DATABASE_URL = { description = "...", ref = { item = "db", field = "password" }, providers = ["prod_op"] }
+  ```
+
+  `item` is the store's own name for the secret (1Password item title, Vault
+  KV path, AWS secret name or ARN, `.env` key, environment variable, ...);
+  optional keys refine it where the store supports them: `field` (1Password
+  field label, Vault KV field, AWS JSON key, keyring account), `vault` and
+  `section` (1Password), and `version` (Google Secret Manager). Every provider
+  resolves refs; coordinates a store has no equivalent for are rejected with a
+  clear error rather than guessed at.
+
+  The coordinates supply naming only — *which* store resolves them follows the
+  same routing as every other secret (the secret's `providers` chain, the
+  `--provider`/`SECRETSPEC_PROVIDER` override, or the default provider). That
+  means refs compose with `providers` fallback chains, and an explicit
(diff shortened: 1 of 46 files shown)


Masqueey:
Addresses https://github.com/cachix/secretspec/issues/123

PR #122, opened 2026-07-06
feat(cli): prompt in a GUI instead of the console, plus a global approval policy

domenkozar:
## Summary

Trusted local prompts for **value entry** and **secret release**, so a CI job or coding agent can *trigger* secret operations without ever seeing the values, and cannot self-approve.

- Mark a secret `interactive = true` (or `secretspec set --ask`) to type its value into a trusted dialog instead of stdin.
- Set `require_approval` in `[project]` (`true` / `false` / `"agents"`) to require human approval at that same trusted dialog before `secretspec run` injects secrets or `secretspec get`/`import` releases one.
- Because the prompt is read on a channel the caller does not control, the triggering tool cannot forge or self-answer it.

## Trusted channel: a built-in GUI

The CLI renders the prompt with a **self-contained GUI dialog** and no longer needs an external `pinentry` binary installed. The channel resolves as:

1. **Built-in `egui-pinentry` dialog** (with the `gui-prompt` feature, enabled by `cli`) — a small window drawn with egui **on the CPU** (no GPU / OpenGL / Vulkan), that on X11 **grabs the keyboard** to resist snooping.
2. A GnuPG `pinentry` binary when built without `gui-prompt`.
3. A `/dev/tty` prompt when no display is available. A detected agent is refused any terminal-bound channel.

## New crate: `egui-pinentry`

A self-contained passphrase + confirmation dialog crate (its own commit; publishable independently):

- Pure-CPU render via `softbuffer` + `egui_software_backend` (no GPU crates in the tree).
- X11 `XGrabKeyboard`, released via a `Drop` guard; no-op on Wayland/macOS/Windows.
- Keystrokes go into a `zeroize::Zeroizing` buffer (never an egui `TextEdit`), so the secret is not retained in egui's widget/undo state; returned as a `SecretString`.
- One event loop reused across dialogs (`run_app_on_demand`).
- Headless UI tests with `egui_kittest`.

## SDKs stay lean

`secretspec-ffi`, `-py`, and `-node` now depend on the library with `default-features = false` (+ the providers they resolve from), so `gui-prompt` and the ~78 GUI crates are **never** pulled into the language SDK cdylibs. Verified: building the ffi crate compiles egui-pinentry 0×, the CLI 1×.

## Notes

- 333 unit tests pass (trusted-prompt/approval paths use test hooks that bypass the real window).
- An `interactive` secret (or `set --ask`) never reads a piped value: `echo v | secretspec set KEY` prompts rather than storing `v`. Pass it inline (`secretspec set KEY "$V"`) or leave the secret non-`interactive`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,21 +8,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **Trusted prompts for value entry and secret release.** Keep secrets off an
-  orchestrator's pipes so a CI job or coding agent can *trigger* secret operations
-  without ever seeing the values. Mark a secret `interactive = true` (or pass
-  `secretspec set --ask`) to type its value into a trusted local dialog instead
-  of stdin. The CLI shows a built-in GUI dialog that needs no external program
-  (it grabs the keyboard on X11), falling back to a `/dev/tty` prompt when no
-  display is available. Set `require_approval` in `[project]`
-  (`true`, `false`, or `"agents"`) to require human approval at that same trusted
-  prompt before `secretspec run` injects secrets into a command or `secretspec get`
-  prints one. Because the prompt is read on a channel the caller does not control,
-  the triggering tool cannot self-approve. Works with every provider. Note that an
-  `interactive` secret (or `set --ask`) never reads a piped value: `echo v |
-  secretspec set KEY` prompts rather than storing `v`. To set from a script or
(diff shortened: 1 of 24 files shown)


domenkozar:
Blocked on https://github.com/dsociative/pinentry-egui/issues/1

domenkozar:
global policy should also include required approval for only certain secrets, e.g. production profile

domenkozar:
I'm closing this for now and #300 is the general place where we have to get this right first.

PR #127, opened 2026-07-10
feat(provider): add Passbolt provider via go-passbolt-cli

Tails:
Add a `passbolt://` provider backed by the official go-passbolt-cli, for the self-hosted Passbolt password manager. Convention secrets map to a resource named `secretspec/{project}/{profile}/{key}` with the value in the `password` field; a secret's `ref` can instead point at an existing resource by id or name and pick a field (password/username/uri/description).

Credentials come either from the CLI's own configuration or from secretspec-owned env vars (SECRETSPEC_PASSBOLT_SERVER, SECRETSPEC_PASSBOLT_PRIVATE_KEY_FILE / SECRETSPEC_PASSBOLT_PRIVATE_KEY, SECRETSPEC_PASSBOLT_PASSPHRASE) that the provider forwards to the CLI — the passphrase and inline key via the child's environment, never the argv — so no credentials live in secretspec.toml or the provider URI.

Includes docs (provider page, reference, concepts, quick-start) and a CHANGELOG entry.

### DIFF
--- a/docs/astro.config.ts
+++ b/docs/astro.config.ts
@@ -168,6 +168,7 @@ Secrets can be stored in: keyring (default), dotenv files, environment variables
             { label: "Environment Variables", slug: "providers/env" },
             { label: "Pass", slug: "providers/pass" },
             { label: "Proton Pass", slug: "providers/protonpass" },
+            { label: "Passbolt", slug: "providers/passbolt" },
             { label: "LastPass", slug: "providers/lastpass" },
             { label: "1Password", slug: "providers/onepassword" },
             {
--- a/docs/src/content/docs/providers/passbolt.md
+++ b/docs/src/content/docs/providers/passbolt.md
@@ -1,0 +1,147 @@
+---
+title: Passbolt Provider
+description: Passbolt self-hosted password manager via go-passbolt-cli
+---
+
+The Passbolt provider integrates with [Passbolt](https://www.passbolt.com/), the
+self-hosted, open-source password manager, via the official
+[`go-passbolt-cli`](https://github.com/passbolt/go-passbolt-cli) (`passbolt`).
+
+It fits the workflow where **a human enters an API key or service password once**
+(via the Passbolt web UI or CLI) and **dev machines read it at runtime** — no
+secret material is ever written to `secretspec.toml` or to disk.
+
+## Prerequisites
+
+- The `passbolt` CLI installed and on `PATH` (build from
+  [go-passbolt-cli](https://github.com/passbolt/go-passbolt-cli)).
+- Credentials for your Passbolt account: the server address, your OpenPGP
+  private key, and its passphrase.
+
+## Authentication
+
+The private key and passphrase are the bootstrap secrets that unlock every
+other secret, so they never belong in `secretspec.toml` or the provider URI.
+There are two ways to supply them.
+
+### Option A — secretspec-owned env vars (no separate CLI config)
+
+Set these in your environment (e.g. via your shell profile, a `.envrc`, or your
+CI secret store) and secretspec forwards them to the CLI for you — the
+passphrase and inline key go through the child process environment, never the
+command line:
+
+| Variable | Purpose |
+|----------|---------|
+| `SECRETSPEC_PASSBOLT_SERVER` | Server address (or use the URI's `?server=`) |
+| `SECRETSPEC_PASSBOLT_PRIVATE_KEY_FILE` | Path to your OpenPGP private key file |
+| `SECRETSPEC_PASSBOLT_PRIVATE_KEY` | Inline OpenPGP private key (alternative to the file) |
+| `SECRETSPEC_PASSBOLT_PASSPHRASE` | Private-key passphrase |
+
+```bash
+export SECRETSPEC_PASSBOLT_SERVER=https://pass.example.com
+export SECRETSPEC_PASSBOLT_PRIVATE_KEY_FILE=~/.config/passbolt/ada.asc
+export SECRETSPEC_PASSBOLT_PASSPHRASE='<passphrase>'
+
+secretspec get STRIPE_SECRET_KEY --provider passbolt
+```
+
+### Option B — the CLI's own configuration
+
+Run `passbolt configure` once; with none of the env vars above set, the
+provider inherits that configuration:
+
+```bash
+passbolt configure \
+  --serverAddress https://pass.example.com \
+  --userPrivateKeyFile ada.asc \
+  --userPassword '<passphrase>'
+```
+
+Either way, no credentials appear in `secretspec.toml` or the provider URI.
+
+## Configuration
+
+### URI Format
+
+```
+passbolt://[name-template][?folder=<folder-id>&server=<server-address>]
+```
+
+- `name-template`: resource-name pattern for convention secrets, supporting the
+  `{project}`, `{profile}`, `{key}` placeholders. Defaults to
+  `secretspec/{project}/{profile}/{key}`.
+- `folder` (optional): a Passbolt folder id. New convention resources are
+  created inside it (`--folderParentID`) and name lookups are scoped to it.
+- `server` (optional): overrides the CLI's configured server address
+  (`--serverAddress`). Useful when one machine talks to several Passbolt servers.
+
+### Examples
+
+```bash
+# Default resource name (secretspec/{project}/{profile}/{key})
+passbolt://
+
+# Custom resource-name template
+passbolt://secretspec/{project}/{profile}/{key}
+
+# Pin the server address (e.g. a tailnet-internal Passbolt)
+passbolt://?server=https://pass.example.com
+
+# Scope convention resources to a folder
+passbolt://?folder=a9230ec4-5507-4870-b8b5-b3f500587e4c
+```
+
+## Storage model
+
+Each convention secret maps to one Passbolt **resource**:
+
+- The resource **name** encodes `{project}/{profile}/{key}`.
+- The secret value lives in the resource's **password** field.
+
+## Secret references
+
+A secret's [`ref`](/reference/configuration/#secret-references) points at an
+existing, human-provisioned resource instead of the convention layout — the
+common case for "someone added the API key in Passbolt, I just want to read it":
+
+- `item`: the resource **id** (a UUID, copied from the Passbolt UI) or the exact
+  resource **name**.
+- `field` (optional): which resource field holds the value — one of `password`
+  (default), `username`, `uri`, `description`.
+
+```toml
+[profiles.production]
+# By resource id (unambiguous — recommended for refs)
+STRIPE_SECRET_KEY = { description = "Stripe key", ref = { item = "a9230ec4-5507-4870-b8b5-b3f500587e4c" }, providers = ["passbolt://?server=https://pass.example.com"] }
+
+# By resource name, reading a non-default field
+SERVICE_USER = { description = "Service account user", ref = { item = "Payments service account", field = "username" }, providers = ["passbolt"] }
+```
+
+Reads fetch the resource by id (or resolve the name to an id first) and decrypt
+the requested field. Writes update that field on the existing resource; a `ref`
+whose id resolves to nothing is an error (secretspec never creates a detached
+resource for an externally managed reference).
+
+## Usage
+
+```bash
+# Set a secret (creates or updates the secretspec/... resource's password)
+$ secretspec set STRIPE_SECRET_KEY --provider passbolt
+Enter value for STRIPE_SECRET_KEY: sk_live_...
+
+# Get a secret
+$ secretspec get STRIPE_SECRET_KEY --provider passbolt
+
+# Run a process with secrets injected
+$ secretspec run --provider passbolt -- npm start
+```
+
+## Environment variables
+
+- `SECRETSPEC_PASSBOLT_SERVER`, `SECRETSPEC_PASSBOLT_PRIVATE_KEY_FILE`,
+  `SECRETSPEC_PASSBOLT_PRIVATE_KEY`, `SECRETSPEC_PASSBOLT_PASSPHRASE`: credentials
+  forwarded to the CLI (see [Authentication](#authentication)).
+- `SECRETSPEC_PASSBOLT_CLI_PATH`: path to the `passbolt` binary when it is not
+  simply `passbolt` on `PATH`.
--- a/secretspec/src/provider/passbolt.rs
+++ b/secretspec/src/provider/passbolt.rs
@@ -1,0 +1,721 @@
+use crate::provider::{Address, Provider, ProviderUrl};
+use crate::{Result, SecretSpecError};
+use secrecy::{ExposeSecret, SecretString};
+use serde::Deserialize;
+use std::io;
+use std::process::{Command, Stdio};
+
+/// The resource field that carries a secret's value when a `ref` does not name
+/// one explicitly. Passbolt resources model a login (name/username/uri/password
+/// /description); the password is the secret material, so it is the default.
+const DEFAULT_FIELD: &str = "password";
+
+/// Resource fields a native `ref` may address via its `field` coordinate. These
+/// are the standard Passbolt resource attributes; `password` is the secret, the
+/// rest are metadata that a `ref` can still pin (e.g. reading a stored `uri`).
+const KNOWN_FIELDS: &[&str] = &["password", "username", "uri", "description", "name"];
+
+/// Environment variables that let secretspec supply the `passbolt` CLI's own
+/// credentials, so `secretspec.toml` stays secret-free and no separate
+/// `passbolt configure` step is needed. All are optional; when none are set the
+/// provider falls back to whatever the CLI is already configured with (its
+/// config file or native env vars).
+///
+/// The passphrase and inline key are *secrets*: they are forwarded to the child
+/// through its environment (the names `go-passbolt-cli`'s viper reads), never on
+/// the argv (which is world-visible via `ps`) and never in the provider URI.
+const ENV_SERVER: &str = "SECRETSPEC_PASSBOLT_SERVER";
+const ENV_PRIVATE_KEY_FILE: &str = "SECRETSPEC_PASSBOLT_PRIVATE_KEY_FILE";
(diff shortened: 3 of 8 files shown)


domenkozar (commented):
Thanks for the contribution! I ran a deep review of this PR (multi-agent, each finding independently verified against the code, the go-passbolt-cli/go-passbolt sources, and where possible the v0.5.1 binary). Inline comments below, roughly ordered by severity in each file.

One finding has no diff line to anchor on, so it goes here:

**docs: `reference/configuration.md` secret-references tables not extended.** The Secret References section of `docs/src/content/docs/reference/configuration.md` has a store-shape table and a per-provider "How providers interpret the coordinates" table listing every ref-capable provider. Passbolt is missing from both, so a user consulting the page that `passbolt.md` itself links to for `ref` semantics would conclude refs are unsupported.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+        Ok(resource
+            .field(field)
+            .map(|v| SecretString::new(v.into())))
+    }
+
+    fn set(&self, addr: Address<'_>, value: &SecretString) -> Result<()> {
+        let coords = self.resolve_coords(addr)?;
+        let field = validate_field(coords.field.as_deref().unwrap_or(DEFAULT_FIELD))?;
+        let flag = Self::field_flag(field);
+        let secret = value.expose_secret();
```
domenkozar, line 458: **security:** `set()` puts the secret value itself on the child argv: `--password <secret>` on both the update path (line 463) and the create path (line 478). The value is world readable via `ps` / `/proc/<pid>/cmdline` for the full lifetime of the CLI call, which spans a complete server login. The pass, lastpass, and protonpass providers pipe secret values via stdin for exactly this reason (lastpass.rs documents it explicitly).

go-passbolt-cli currently accepts the value only as a flag (no stdin/env channel in `internal/cmd/resource` create/update), so this may be unavoidable today. In that case it should at least be a documented limitation: the module docs and provider page currently advertise argv hygiene ("via the child's environment, never the argv") that covers the auth passphrase but not the secret value itself.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+        // externally managed resource that must exist, so we refuse rather than
+        // silently create a detached one.
+        if is_uuid(&coords.item) {
+            return Err(SecretSpecError::ProviderOperationFailed(format!(
+                "Passbolt resource '{}' does not exist; a ref must name an existing resource",
+                coords.item
+            )));
+        }
+
+        let mut args = vec!["create", "resource", "--name", &coords.item, &flag, secret];
```
domenkozar, line 478: **correctness:** a name-based `ref` whose name resolves to nothing falls through to this create path and silently creates a brand new detached resource. The `is_uuid` guard above only protects id refs, but the docs page promises "secretspec never creates a detached resource for an externally managed reference", and its own example uses a name ref (`item = "Payments service account"`).

Concretely: the referenced resource gets renamed or isn't shared with this user yet, then `secretspec set` creates an orphan duplicate instead of erroring. With `field = "username"` / `"uri"` / `"description"` it's worse: the create runs without `--password` and go-passbolt-cli fails with the unrelated "required flag \"password\" not set".

At this point in `set()` a convention name and a user-authored ref name are indistinguishable, so the fix is to decide before `resolve_coords` collapses them, e.g. match on `Address::Native` like onepassword.rs does (its refs never create, for id or name).

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+
+/// The resource field that carries a secret's value when a `ref` does not name
+/// one explicitly. Passbolt resources model a login (name/username/uri/password
+/// /description); the password is the secret material, so it is the default.
+const DEFAULT_FIELD: &str = "password";
+
+/// Resource fields a native `ref` may address via its `field` coordinate. These
+/// are the standard Passbolt resource attributes; `password` is the secret, the
+/// rest are metadata that a `ref` can still pin (e.g. reading a stored `uri`).
+const KNOWN_FIELDS: &[&str] = &["password", "username", "uri", "description", "name"];
```
domenkozar, line 16: **security:** `KNOWN_FIELDS` includes `"name"`, but every doc surface (provider page, reference, CHANGELOG, the rustdoc below) limits `field` to password/username/uri/description. Nothing upstream validates coordinate values, so `ref = { item = "...", field = "name" }` passes `validate_field` and makes `set()` run `passbolt update resource --id X --name <secret>`, renaming the resource to the secret value. For v4 resource types (still the common default) names are cleartext server-side metadata, so that's a secret leak, and it breaks any subsequent lookup by the old name either way. On the create path the duplicate `--name` still puts the secret on the argv, though the CLI then fails its required-password check before creating anything.

Dropping `"name"` from `KNOWN_FIELDS` fixes the write hazard; `find_id_by_name` only needs the struct field, not the coordinate.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+        )))
+    }
+}
+
+/// Recognizes the CLI's "resource does not exist" failures so a missing secret
+/// resolves to `Ok(None)` rather than an error. Passbolt surfaces this as an
+/// HTTP 404 wrapped in the `getting resource:` context.
+fn is_not_found(msg: &str) -> bool {
+    let lower = msg.to_lowercase();
+    lower.contains("404")
```
domenkozar, line 379: **correctness:** matching `"404"` / `"not found"` / `"does not exist"` / `"could not find"` anywhere in the lowercased stderr is too broad for the one place that converts errors into `Ok(None)` (`get_resource`). go-passbolt formats every API error as `API error (code %d): %s`, so a wrong `?server=` base path that 404s, a revoked share (the Passbolt API intentionally answers 404 with "The resource does not exist." for unauthorized resources), and go-passbolt's sentinel errors `resource type not found` / `metadata key not found` all get masked as "secret not set" (the masking hits id-addressed lookups; name lookups error loudly in `find_id_by_name`). `secretspec check` then prompts the user to re-enter values against a broken or unauthorized setup instead of surfacing the real error.

The in-tree precision bar is exact tool phrases (onepassword.rs matches "isn't an item" / "doesn't have a field"); matching the specific composed shape here (e.g. `getting resource` plus code 404) would keep unrelated failures loud.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+    fn set(&self, addr: Address<'_>, value: &SecretString) -> Result<()> {
+        let coords = self.resolve_coords(addr)?;
+        let field = validate_field(coords.field.as_deref().unwrap_or(DEFAULT_FIELD))?;
+        let flag = Self::field_flag(field);
+        let secret = value.expose_secret();
+
+        // Update the resource in place when it already exists (by id or by
+        // resolved name); this is the path a human-provisioned resource takes.
+        if let Some(id) = self.resolve_id(&coords.item)? {
+            self.run(&["update", "resource", "--id", &id, &flag, secret])?;
```
domenkozar, line 463: **correctness:** setting an empty value on an existing resource is a silent success no-op. go-passbolt's `helper.UpdateResource` skips empty-string fields by design ("Empty strings are not applied (partial update)"), so the CLI exits 0 while keeping the old value, and `set()` reports success without persisting anything.

secretspec's CLI `set` path rejects empty values before reaching the provider, but two flows deliver an empty string today: `secretspec check`'s interactive prompt (no empty guard before `backend.set()`) and `import`. Combined with the read side filtering empty fields to `None` (line 57), a resource whose password is empty makes `check` never converge: it reports the secret missing, prompts, saves a no-op, and the re-validation in the same run (and every later run) reports it missing again. Worth rejecting empty values here (or documenting the CLI limitation) so `set` can't return `Ok` without writing.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+            return Ok(());
+        }
+
+        // No such resource: create one. Only a name-addressed convention secret
+        // reaches here — an id (`ref`) that resolves to nothing points at an
+        // externally managed resource that must exist, so we refuse rather than
+        // silently create a detached one.
+        if is_uuid(&coords.item) {
```
domenkozar, line 471: **correctness:** this branch is unreachable. `resolve_id()` returns `Ok(Some(..))` unconditionally for any UUID-shaped item, without an existence check, so reaching this line requires `resolve_id() == None` and `is_uuid() == true` at the same time, which is a contradiction. A dangling UUID ref therefore always takes the update path above and surfaces go-passbolt-cli's raw stderr instead of this crafted message.

Making the message real requires probing existence for UUID items (e.g. via `get_resource`) before choosing between update and refusal.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+        };
+        let Some(resource) = self.get_resource(&id)? else {
+            return Ok(None);
+        };
+        Ok(resource
+            .field(field)
+            .map(|v| SecretString::new(v.into())))
+    }
+
+    fn set(&self, addr: Address<'_>, value: &SecretString) -> Result<()> {
```
domenkozar, line 454: **correctness:** `set()` neither calls `check_writable()` nor overrides it, so the provider inherits the default `Ok(())`. The trait docs say callers use `check_writable` to refuse a write before prompting for a value, and `secrets.rs` does exactly that before showing the password prompt. As written, a doomed write (e.g. a dangling ref) passes the pre-prompt gate, the user types the secret, and only then does `set()` fail, discarding what they typed. vault.rs, awssm.rs, and gcsm.rs override it and have `set()` call it; same pattern would fit here once the ref-existence policy from the comment above is decided.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            if stderr.contains("is not defined") || stderr.contains("serverAddress") {
+                return Err(SecretSpecError::ProviderOperationFailed(format!(
+                    "Passbolt CLI is not configured. Run 'passbolt configure --serverAddress \
+                     https://... --userPrivateKeyFile key.asc --userPassword <passphrase>' \
+                     (details: {})",
+                    stderr.trim()
+                )));
+            }
+            return Err(SecretSpecError::ProviderOperationFailed(stderr.to_string()));
```
domenkozar, line 301: **error handling:** when the passphrase is missing (e.g. `SECRETSPEC_PASSBOLT_PASSPHRASE` unset and `passbolt configure` was run without `--userPassword`), the CLI tries to prompt, but `Command::output()` closes the child's stdin, so `ReadPassword` gets EOF and the user sees the raw `Error: reading Password: EOF` (verified against the v0.5.1 binary). That stderr matches neither heuristic in `run()`, so there's no pointer to `SECRETSPEC_PASSBOLT_PASSPHRASE` even though the provider defines it. Same class of failure for MFA accounts: the default `mfaMode` is interactive-totp, so they get `Error: reading TOTP: EOF` even with a passphrase supplied.

Mapping these prompt-EOF failures to an actionable message (mention the env var, and `--mfaMode noninteractive-totp` for MFA) would cover the most likely first-run failure.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+        if let Some(folder) = &self.config.folder_id {
+            args.push("--folder");
+            args.push(folder);
+        }
+        let output = self.run(&args)?;
+        let resources: Vec<PassboltResource> = serde_json::from_str(&output)
+            .map_err(|e| SecretSpecError::ProviderOperationFailed(e.to_string()))?;
+        Ok(resources
+            .into_iter()
+            .find(|r| r.name.as_deref() == Some(name))
```
domenkozar, line 325: **correctness:** Passbolt allows multiple resources with the same name, and this silently takes the first match. The server's default listing order sorts by name, so same-name duplicates tie on the sort key and their relative order is an unspecified database tie-break: which copy wins is arbitrary, and reads or writes can target a different copy than the one the user provisioned (e.g. a teammate's shared duplicate vs their own), with the write appearing lost and no error ever raised.

Detecting duplicates here and erroring (or at least warning) would make the ambiguity visible.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+            if !uri.contains("://") {
+                uri.push_str("://");
+            }
+            uri.push('?');
+            uri.push_str(&query.join("&"));
+        }
+        uri
+    }
+
+    fn get(&self, addr: Address<'_>) -> Result<Option<SecretString>> {
```
domenkozar, line 439: **efficiency:** without a `get_many` override, every convention secret pays a full-account `passbolt list resource` (each CLI spawn redoes the whole login) plus a `get resource`. `secretspec run` with N secrets fires N identical account-wide listings concurrently, so 20 secrets against an account with a few thousand resources means 40 process spawns, 40 full logins, and 19 redundant listings (folder-scoped when `?folder=` is set). onepassword, protonpass, bws, and awssm all override `get_many` with one listing plus parallel fetches for this reason.

A `get_many` here could also serve `username`/`uri` ref fields straight from the listing (those columns don't require secret decryption), keeping the second spawn only for password/description.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+///
+/// [`ref`]: crate::config::NativeAddress
+pub struct PassboltProvider {
+    config: PassboltConfig,
+    /// Path to the `passbolt` binary. Override with the
+    /// `SECRETSPEC_PASSBOLT_CLI_PATH` environment variable.
+    cli_binary_path: String,
+}
+
+crate::register_provider! {
```
domenkozar, line 200: **consistency:** the other remote CLI providers (onepassword, protonpass, lastpass) register a `preflight:` probe in `register_provider!` plus an `auth_scope_key` override, so a locked or misconfigured CLI fails once with one clear error before any per-secret work. Without them, a bad passphrase and 10 secrets means 10 concurrent CLI processes each attempting a full login: 10 duplicate raw errors and 10 wasted logins, plus noise for any rate limiting or fail2ban rules in front of the server. A cheap probe wired into `preflight` would match the sibling behavior.

secretspec/src/provider/passbolt.rs
```diff
@@ -0,0 +1,721 @@
...
+    fn find_id_by_name(&self, name: &str) -> Result<Option<String>> {
+        let mut args = vec![
+            "list", "resource", "--json", "--column", "id", "--column", "name",
+        ];
+        if let Some(folder) = &self.config.folder_id {
+            args.push("--folder");
+            args.push(folder);
+        }
+        let output = self.run(&args)?;
+        let resources: Vec<PassboltResource> = serde_json::from_str(&output)
```
domenkozar, line 321: **nit/reuse:** `SecretSpecError` already has `Json(#[from] serde_json::Error)`, so this `map_err` (and the one in `get_resource` below) can be plain `?`, like onepassword.rs's item decode. That also classifies these failures as `json` for `SecretSpecError::kind()` instead of `provider_operation_failed`.

docs/src/content/docs/providers/passbolt.md
```diff
@@ -0,0 +1,147 @@
...
+- `field` (optional): which resource field holds the value — one of `password`
+  (default), `username`, `uri`, `description`.
+
+```toml
+[profiles.production]
+# By resource id (unambiguous — recommended for refs)
+STRIPE_SECRET_KEY = { description = "Stripe key", ref = { item = "a9230ec4-5507-4870-b8b5-b3f500587e4c" }, providers = ["passbolt://?server=https://pass.example.com"] }
+
+# By resource name, reading a non-default field
+SERVICE_USER = { description = "Service account user", ref = { item = "Payments service account", field = "username" }, providers = ["passbolt"] }
```
domenkozar, line 119: **docs:** this example fails at runtime. A bare name in a per-secret `providers` chain must be a declared alias (project `[providers]` table or user config), not a provider name; only full `scheme://` URIs pass through directly. Copying this line as-is yields "Provider alias 'passbolt' is not defined". Either use `providers = ["passbolt://"]` like the STRIPE example above effectively does, or add the `[providers]` alias table to the example.

docs/astro.config.ts
```diff
@@ -168,6 +168,7 @@ Secrets can be stored in: keyring (default), dotenv files, environment variables
             { label: "Environment Variables", slug: "providers/env" },
             { label: "Pass", slug: "providers/pass" },
             { label: "Proton Pass", slug: "providers/protonpass" },
+            { label: "Passbolt", slug: "providers/passbolt" },
```
domenkozar, line 171: **docs drift:** the sidebar entry is here, but the other provider listings from the CLAUDE.md "Adding Provider Documentation" checklist were missed:

- the providers sentence in the `starlightLlmsTxt` description block in this file (still ends at "Bitwarden Secrets Manager")
- `docs/src/pages/index.astro`: the `providerMetadata` array has no passbolt entry, and the page's "11 backends" counts and provider enumerations are now stale
- `README.md`: the Providers bullet list and the `secretspec config init` example output (quick-start.mdx got the new line, the README didn't)

After this lands, llms.txt, the landing page, and the GitHub README all advertise a provider list that no longer matches what the binary prints.

Tails: Excellent work, thanks. I had done the pull request to document the possibility of using secretspec for Passbolt, but I'll take up the challenge and go through these implementation drawbacks and make it proper. 

And many thanks for secretspec, it's awesome. Have been using it for every project since becoming aware of it!

domenkozar:
It needs to be also reworked for provider credentials in https://secretspec.dev/concepts/providers/#provider-credentials

domenkozar:
@Tails ping :)

domenkozar:
@Tails I'm planning to do a 0.17 release on Monday if you'd like to get this in.

Tails:
I was rebasing on 0.16, but I'm on holiday, so feel free to release and I'll rerebase on that.

domenkozar:
@Tails 0.18 targeting Monday

domenkozar:
@Tails do you mind me taking over? 0.19 is scheduled for Monday

domenkozar:
I've preserved your authorship and made #301

PR #132, opened 2026-07-13
Add Azure Key Vault provider (akv://)

phoem:
Adds an akv:// provider backed by Azure Key Vault, gated behind a new akv Cargo feature (included in default features alongside gcsm/awssm/vault/bws). Convention secrets map to Azure Key Vault's [0-9a-zA-Z-] secret-name charset via secretspec--{project}--{profile}--{key}, with underscores rewritten to hyphens. Authentication is chosen via ?auth=env|cli|managed_identity|workload_identity, defaulting to a service principal from AZURE_TENANT_ID/AZURE_CLIENT_ID/AZURE_CLIENT_SECRET and falling back to the signed-in Azure CLI session.

Version pinning and user-assigned managed identity are not yet supported. Updates docs (new providers/akv.md page plus the other tracked locations) and CHANGELOG.md.

### DIFF
--- a/docs/src/content/docs/providers/akv.md
+++ b/docs/src/content/docs/providers/akv.md
@@ -16,15 +16,16 @@ The Azure Key Vault provider integrates with Azure for centralized secret manage
 ### URI Format
 
 ```
-akv://VAULT_NAME[?auth=env|cli|managed_identity|workload_identity]
+akv://VAULT_NAME[?auth=env|cli|managed_identity|workload_identity][&suffix=DNS_SUFFIX]
 ```
 
 - `VAULT_NAME`: Your Key Vault name (e.g. `myvault`), or a full DNS name for sovereign clouds (e.g. `myvault.vault.azure.cn`)
 - `auth`: Authentication method (default: `env`)
-  - `env` — a service principal from `AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET`, falling back to the signed-in Azure CLI / Azure Developer CLI session if those aren't set
+  - `env` — a service principal from `AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET` (all three must be set together); falls back to the signed-in Azure CLI / Azure Developer CLI session if none are set. Setting only some of the three is an error rather than a silent fallback to a different identity.
   - `cli` — the Azure CLI / Azure Developer CLI session only
   - `managed_identity` — the VM / App Service / AKS system-assigned managed identity
   - `workload_identity` — AKS workload identity federation (`AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_FEDERATED_TOKEN_FILE`, injected automatically by AKS)
+- `suffix`: an explicit Key Vault DNS suffix for a bare `VAULT_NAME`, e.g. `akv://myvault?suffix=vault.azure.cn` for a sovereign cloud, instead of relying on a dotted `VAULT_NAME`
 
 ### Examples
 
@@ -40,6 +41,9 @@ $ secretspec check --provider akv://myvault?auth=managed_identity
 
 # Run with secrets
 $ secretspec run --provider akv://myvault -- npm start
+
+# Sovereign cloud, via an explicit suffix instead of a dotted vault name
+$ secretspec check --provider akv://myvault?suffix=vault.azure.cn
 ```
 
 ## Secret References
@@ -47,7 +51,11 @@ $ secretspec run --provider akv://myvault -- npm start
 By default each secret is stored as `secretspec--{project}--{profile}--{key}`. A
 secret's [`ref`](/reference/configuration/#secret-references) field names an
 existing secret instead: `item` is the secret name (`field` and `version` are
-not yet supported). References are **read-only** in this provider.
+not yet supported). References are **read-only** in this provider, and `item`
+must already be a valid Azure Key Vault secret name (letters, digits, and
+hyphens only) — unlike convention secrets, it is validated but never rewritten,
+since silently rewriting a `ref` could point at a different secret than the
+one you named.
 
 ```toml
 [profiles.production]
@@ -64,8 +72,9 @@ $ secretspec set DATABASE_URL --provider akv://myvault
 Enter value for DATABASE_URL: postgresql://localhost/mydb
 ✓ Secret 'DATABASE_URL' saved to akv (profile: default)
 
-# Import from .env
-$ secretspec import dotenv://.env
+# Get it back
+$ secretspec get DATABASE_URL --provider akv://myvault
+postgresql://localhost/mydb
 ```
 
 ### Secret Naming
@@ -80,7 +89,11 @@ Example: `DATABASE_URL` in project `myapp`, profile `production` becomes
 
 This rewrite is lossy: a project, profile, or key that differs only in using
 hyphens versus underscores (e.g. `FOO_BAR` and `FOO-BAR`) maps to the same Key
-Vault secret name. Prefer hyphens if that distinction matters to you.
+Vault secret name. Prefer hyphens if that distinction matters to you. A
+project, profile, or key containing a literal `--` or a `__` (which sanitizes
+to `--`) is rejected outright — it would be indistinguishable from the `--`
+delimiter between components and could otherwise silently collide with an
+unrelated secret.
 
 ### CI/CD with a Service Principal
 
@@ -93,6 +106,10 @@ $ export AZURE_CLIENT_SECRET="..."
 # Run command
 $ secretspec run --provider akv://myvault -- deploy
 ```
+
+All three environment variables must be set together; setting only some of
+them is treated as a configuration error rather than a silent fallback to the
+Azure CLI session.
 
 ### AKS with Workload Identity
 
--- a/docs/src/content/docs/providers/akv.md
+++ b/docs/src/content/docs/providers/akv.md
@@ -1,0 +1,103 @@
+---
+title: Azure Key Vault Provider
+description: Azure Key Vault integration
+---
+
+The Azure Key Vault provider integrates with Azure for centralized secret management.
+
+## Prerequisites
+
+- An Azure Key Vault instance
+- Authenticated via a service principal (`AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET`), the Azure CLI (`az login`), a managed identity, or AKS workload identity
+- Build with `--features akv`
+
+## Configuration
+
+### URI Format
+
+```
+akv://VAULT_NAME[?auth=env|cli|managed_identity|workload_identity]
+```
+
+- `VAULT_NAME`: Your Key Vault name (e.g. `myvault`), or a full DNS name for sovereign clouds (e.g. `myvault.vault.azure.cn`)
+- `auth`: Authentication method (default: `env`)
+  - `env` — a service principal from `AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET`, falling back to the signed-in Azure CLI / Azure Developer CLI session if those aren't set
+  - `cli` — the Azure CLI / Azure Developer CLI session only
+  - `managed_identity` — the VM / App Service / AKS system-assigned managed identity
+  - `workload_identity` — AKS workload identity federation (`AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_FEDERATED_TOKEN_FILE`, injected automatically by AKS)
+
+### Examples
+
+```bash
+# Set a secret (reads env vars, or falls back to `az login`)
+$ secretspec set DATABASE_URL --provider akv://myvault
+
+# Get a secret
+$ secretspec get DATABASE_URL --provider akv://myvault
+
+# Check secrets using a managed identity
+$ secretspec check --provider akv://myvault?auth=managed_identity
+
+# Run with secrets
+$ secretspec run --provider akv://myvault -- npm start
+```
+
+## Secret References
+
+By default each secret is stored as `secretspec--{project}--{profile}--{key}`. A
+secret's [`ref`](/reference/configuration/#secret-references) field names an
+existing secret instead: `item` is the secret name (`field` and `version` are
+not yet supported). References are **read-only** in this provider.
+
+```toml
+[profiles.production]
+DATABASE_URL = { description = "DB", ref = { item = "database-url" }, providers = ["akv://myvault"] }
+```
+
+## Usage
+
+### Basic Commands
+
+```bash
+# Set a secret
+$ secretspec set DATABASE_URL --provider akv://myvault
+Enter value for DATABASE_URL: postgresql://localhost/mydb
+✓ Secret 'DATABASE_URL' saved to akv (profile: default)
+
+# Import from .env
+$ secretspec import dotenv://.env
+```
+
+### Secret Naming
+
+Azure Key Vault secret names may only contain ASCII letters, digits and
+hyphens — notably, unlike every other cloud provider, not underscores or
+slashes. Secrets are stored as: `secretspec--{project}--{profile}--{key}`,
+with underscores in each component rewritten to hyphens.
+
+Example: `DATABASE_URL` in project `myapp`, profile `production` becomes
+`secretspec--myapp--production--DATABASE-URL`.
+
+This rewrite is lossy: a project, profile, or key that differs only in using
+hyphens versus underscores (e.g. `FOO_BAR` and `FOO-BAR`) maps to the same Key
+Vault secret name. Prefer hyphens if that distinction matters to you.
+
+### CI/CD with a Service Principal
+
+```bash
+# Set credentials
+$ export AZURE_TENANT_ID="..."
+$ export AZURE_CLIENT_ID="..."
+$ export AZURE_CLIENT_SECRET="..."
+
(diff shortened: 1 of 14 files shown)


secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+                let client_secret = std::env::var("AZURE_CLIENT_SECRET").ok();
+
+                if let (Some(tenant_id), Some(client_id), Some(client_secret)) =
+                    (tenant_id, client_id, client_secret)
+                {
+                    // NOTE: `azure_identity` is a very new (v1.0), fast-moving
+                    // crate; verify this constructor's exact parameter types
+                    // with `cargo check --features akv` before relying on it,
+                    // in case the secret parameter type has changed.
+                    Ok(ClientSecretCredential::new(
```
domenkozar: **Blocking — does not compile.** `azure_identity` 1.0.0 is `ClientSecretCredential::new(tenant_id: &str, client_id: String, secret: azure_core::credentials::Secret, options)`. Here `tenant_id` is a `String` (needs `&str`) and the third arg is a `secrecy::SecretString` (needs `azure_core::credentials::Secret`). `cargo check --features akv` → `error[E0308]` (exit 101), and since `akv` is in the default feature set this breaks `cargo build` / `cargo test --all` / CI.

```rust
Ok(ClientSecretCredential::new(
    &tenant_id,
    client_id,
    azure_core::credentials::Secret::from(client_secret),
    None,
)...
```

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+                "Failed to create Azure Key Vault client for {}: {}",
+                self.config.vault_url, e
+            ))
+        })
+    }
+
+    /// Checks whether an error indicates the secret was not found.
+    fn is_not_found_error(e: &impl std::error::Error) -> bool {
+        let s = e.to_string();
+        s.contains("SecretNotFound") || s.contains("404")
```
domenkozar: **Blocking correctness — not-found is not detected.** A Key Vault 404 is built via `Error::with_message(kind, message)` where `message` is the service's human text ("A secret with (name) was not found in this key vault"); the `SecretNotFound` code sits in a separate `error_code` field and `Error`'s `Display` renders only the message. So this `contains("SecretNotFound") || contains("404")` is **false for a genuine 404**, and `get_secret_async` returns `Err` instead of `Ok(None)` — breaking `check`, `run` with optional/default secrets, and provider fallback for any secret not yet in the vault. The `"404"` substring also risks the opposite (swallowing unrelated errors that happen to contain `404`).

`azure_core::Error` exposes a typed accessor — prefer:
```rust
e.http_status() == Some(azure_core::http::StatusCode::NotFound)
```
(which needs `is_not_found_error` to take `&azure_core::Error` rather than `&impl std::error::Error`).

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+                        e
+                    ))
+                })? as Arc<dyn TokenCredential>)
+            }
+            AuthMethod::Env => {
+                let tenant_id = std::env::var("AZURE_TENANT_ID").ok();
+                let client_id = std::env::var("AZURE_CLIENT_ID").ok();
+                let client_secret = std::env::var("AZURE_CLIENT_SECRET").ok();
+
+                if let (Some(tenant_id), Some(client_id), Some(client_secret)) =
```
domenkozar: **Correctness — partial/empty credentials silently change identity.** This requires all three vars; with only two set, the code drops to the CLI/azd session (line ~288) — authenticating as the developer's personal `az login` identity in dev, or failing with a confusing "failed to fall back to the Azure CLI" in CI, instead of "AZURE_CLIENT_SECRET is missing". Also `std::env::var().ok()` returns `Some("")` for a blank `AZURE_CLIENT_SECRET=`, which counts as "set" and blocks the fallback. Suggest treating empty as unset and erroring on a partially-specified principal.

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+    /// Converts the SecretSpec path format to an Azure-compatible name:
+    /// `secretspec--{project}--{profile}--{key}`, with underscores in each
+    /// component rewritten to hyphens (Azure Key Vault secret names may only
+    /// contain `[0-9a-zA-Z-]`, 1-127 characters).
+    fn format_secret_name(project: &str, profile: &str, key: &str) -> Result<String> {
+        Self::validate_name_component("project", project)?;
+        Self::validate_name_component("profile", profile)?;
+        Self::validate_name_component("key", key)?;
+
+        let sanitize = |s: &str| s.replace('_', "-");
```
domenkozar: **Correctness — name mapping is not injective.** `_`→`-` sanitization plus the `--` delimiter means distinct triples collide: `format_secret_name("a","b__c","d")` and `format_secret_name("a","b","c__d")` both produce `secretspec--a--b--c--d`; likewise key `A__B` collides with `A--B`. `__` is common in real keys (e.g. .NET `Section__Key`), so a `set` on one silently versions over the other and `get` cross-reads. The module doc only warns about the single-underscore case. Every other provider separates with `/`, which components can't contain, so `akv` is uniquely exposed — worth at least documenting the `__` case, ideally rejecting it.

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+            };
+            uri.push_str("?auth=");
+            uri.push_str(auth);
+        }
+        uri
+    }
+
+    fn get(&self, addr: Address<'_>) -> Result<Option<SecretString>> {
+        let coords = self.resolve_coords(addr)?;
+        super::block_on(self.get_secret_async(&coords.item))
```
domenkozar: **Correctness (minor) — native `ref` items skip validation.** `resolve_coords` returns a `Native` address's `item` verbatim, so `ref = { item = "my_secret" }` reaches Azure with an underscore (no charset check, no `_`→`-` rewrite) and fails with a raw opaque API error, whereas the convention path validates and rewrites up front. Consider validating native items too (or documenting that refs must already be Azure-legal).

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+                            e
+                        ))
+                    })? as Arc<dyn TokenCredential>)
+                }
+            }
+        }
+    }
+
+    /// Creates a SecretClient for the configured vault.
+    fn create_client(&self) -> Result<SecretClient> {
```
domenkozar, line 428: **Efficiency — client + credential rebuilt every call.** `create_client()` (called from both `get_secret_async` and `set_secret_async`) constructs a fresh `TokenCredential` each time, discarding the SDK's in-instance token cache; and `get_many` is not overridden, so N secrets fan out to N `get()` calls → N credential resolutions. With `auth=cli`/env-fallback that's one `az account get-access-token` subprocess per secret. Caching the client/credential (e.g. a `OnceLock`) would let the fan-out share a single token.

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+    fn resolve_credential(&self) -> Result<Arc<dyn TokenCredential>> {
+        match self.config.auth {
+            AuthMethod::Cli => Ok(DeveloperToolsCredential::new(None).map_err(|e| {
+                SecretSpecError::ProviderOperationFailed(format!(
+                    "Failed to create Azure CLI / azd credential: {}",
+                    e
+                ))
+            })? as Arc<dyn TokenCredential>),
+            AuthMethod::ManagedIdentity => {
+                Ok(ManagedIdentityCredential::new(None).map_err(|e| {
```
domenkozar, line 370: **Capability/doc gap — managed identity is system-assigned only.** `ManagedIdentityCredential::new(None)` cannot bind a user-assigned identity (no `client_id`), yet `docs/src/content/docs/reference/providers.md` advertises "managed identity" without the qualifier. A VM/AKS pod with only a user-assigned identity would get a runtime token failure. Either support a `client_id` param or make the docs say system-assigned (the provider page already does).

domenkozar (commented):
## Review — Azure Key Vault provider (`akv://`)

Nice, well-documented provider — the whole `CLAUDE.md` provider-doc checklist is satisfied and the CHANGELOG entry is present. However, there are **two blocking issues** plus several correctness edges and cleanups. Details are in inline comments; summary below.

### Blocking
1. **Does not compile.** `ClientSecretCredential::new` (azure_identity 1.0.0) is `new(tenant_id: &str, client_id: String, secret: azure_core::credentials::Secret, ...)`, but the code passes `tenant_id: String` and a `secrecy::SecretString`. `cargo check --features akv` fails with E0308 (exit 101). Because `akv` is in the **default** feature set, `cargo build`, `cargo test --all`, and CI all fail. Fix: `&tenant_id` and `azure_core::credentials::Secret::from(client_secret)`.
2. **`is_not_found_error` misses real 404s.** A Key Vault 404's `Display` is only the service message ("A secret … was not found in this key vault"); the `SecretNotFound` code lives in a separate `error_code` field, so `contains("SecretNotFound") || contains("404")` is false for a genuine not-found and `get` returns `Err` instead of `Ok(None)`. That breaks `check`, `run` with optional/default secrets, and provider fallback chains. Use the typed `e.http_status() == Some(StatusCode::NotFound)`.

### Correctness (moderate)
3. Partial/empty `AZURE_*` env vars silently fall back to a different identity instead of erroring.
4. Secret-name mapping is non-injective (`A__B` and `A--B` collide onto one vault entry).
5. Native `ref` items skip the charset validation/rewrite that convention names get.

### Design / cleanup
6. A fresh client + credential is built per `get`/`set` (no token caching, no `get_many` override) → N token fetches / `az` spawns for N secrets.
7. Managed identity is system-assigned only, but the reference docs advertise it unqualified.
8. Sovereign cloud is selected by a `contains('.')` heuristic — a bare name silently hits the public endpoint.
9. `version` coordinate is rejected though Key Vault versions secrets (known limitation).
10. Leftover dev `NOTE` comment.
11. `uri()` duplicates the auth-method↔string map.

### Test / docs (non-blocking, no inline anchor)
12. `secretspec/src/provider/tests.rs` (~line 482): the generic integration harness has no `akv` arm, so `SECRETSPEC_TEST_PROVIDERS=akv` falls into the `_` branch, calls `try_from("akv")` (bare name has no vault), and panics with a misleading "akv provider should exist". akv gets zero coverage in the shared suite. (This also confirms CI never reaches the Azure network.)

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+
+        let vault_host = url.host().filter(|s| !s.is_empty()).ok_or_else(|| {
+            SecretSpecError::ProviderOperationFailed(
+                "Azure Key Vault name is required. Use format: akv://myvault".to_string(),
+            )
+        })?;
+
+        // A host containing a dot is already a full DNS name (sovereign
+        // clouds); a bare name gets the public-cloud suffix appended.
+        let vault_url = if vault_host.contains('.') {
```
domenkozar: **Altitude — sovereign cloud picked by a magic heuristic.** `vault_host.contains('.')` means a bare vault name in China/USGov/Germany silently gets the public `.vault.azure.net` suffix (wrong endpoint, failing auth/DNS with no hint), and a trailing-dot typo `akv://myvault.` becomes `https://myvault./`. An explicit `?cloud=`/`?suffix=` param (mirroring `?auth=`) would be a first-class, testable knob rather than an inferred special case.

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+    /// Checks whether an error indicates the secret was not found.
+    fn is_not_found_error(e: &impl std::error::Error) -> bool {
+        let s = e.to_string();
+        s.contains("SecretNotFound") || s.contains("404")
+    }
+
+    /// Retrieves a secret's current value by name, mapping "not found" to `None`.
+    async fn get_secret_async(&self, name: &str) -> Result<Option<SecretString>> {
+        let client = self.create_client()?;
+        match client.get_secret(name, None).await {
```
domenkozar, line 451: **Feature note — no version pinning.** `get_secret(name, None)` hardcodes the latest version and `supported_coords` stays empty, so `ref = { version = "…" }` is rejected even though Key Vault genuinely versions secrets. Documented as a known limitation and safe-by-default (rejects rather than silently ignores), just flagging for the roadmap.

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+            }
+            AuthMethod::Env => {
+                let tenant_id = std::env::var("AZURE_TENANT_ID").ok();
+                let client_id = std::env::var("AZURE_CLIENT_ID").ok();
+                let client_secret = std::env::var("AZURE_CLIENT_SECRET").ok();
+
+                if let (Some(tenant_id), Some(client_id), Some(client_secret)) =
+                    (tenant_id, client_id, client_secret)
+                {
+                    // NOTE: `azure_identity` is a very new (v1.0), fast-moving
```
domenkozar: **Cleanup — leftover dev note.** This `NOTE:` block asks the reader to verify the constructor with `cargo check` "in case the secret parameter type has changed" — authoring uncertainty rather than documentation, and it happens to guard the exact call that fails to compile. Please drop it (keep the legitimate "fall back to CLI" comment below).

secretspec/src/provider/akv.rs
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    fn name(&self) -> &'static str {
+        Self::PROVIDER_NAME
+    }
+
+    fn uri(&self) -> String {
+        let mut uri = format!("akv://{}", self.config.vault_host);
+        if self.config.auth != AuthMethod::default() {
+            let auth = match self.config.auth {
```
domenkozar: **Cleanup — duplicated auth mapping.** This `AuthMethod`→`&str` match is the exact inverse of the `&str`→`AuthMethod` match in `TryFrom` (lines ~124-134). Two matches that must stay in lockstep with no shared source of truth — add/rename an auth method and miss one, and `uri()` emits a string `TryFrom` can no longer parse. Consider a single `AuthMethod::as_str()` / `parse` pair.

docs/src/content/docs/providers/akv.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Basic Commands
+
+```bash
+# Set a secret
+$ secretspec set DATABASE_URL --provider akv://myvault
+Enter value for DATABASE_URL: postgresql://localhost/mydb
+✓ Secret 'DATABASE_URL' saved to akv (profile: default)
+
+# Import from .env
```
domenkozar: **Docs — misleading example.** `secretspec import dotenv://.env` imports *from* a local `.env` into the default provider; it has nothing to do with Azure Key Vault, so it reads oddly on the akv page (a reader might expect it to load secrets *into* the vault). The command is valid, just out of place here.

phoem:
My apologies I did not mean for this to be active yet. I am making fixes/changes and I will let you know when they're all resolved. Thanks also for the great and detailed comments.

secretspec/src/provider/akv.rs
phoem: Thanks, fixed.  I somehow set the PR as ready ahead of it being ready.  Apologies, appreciate the great comments though.

secretspec/src/provider/akv.rs
phoem: Removed the comment.

secretspec/src/provider/akv.rs
phoem: Fixed — now checks e.http_status() == Some(StatusCode::NotFound) on the typed error instead of string matching.

secretspec/src/provider/akv.rs
phoem: Fixed — blank vars are now treated as unset, and a partially-specified principal now errors (naming the missing var) instead of silently falling back to the CLI session.

secretspec/src/provider/akv.rs
phoem: Went further than documenting it — now rejects any project/profile/key component containing -- or __ before it can collide with the delimiter.

secretspec/src/provider/akv.rs
phoem: Fixed — native ref items are now validated against Azure's charset up front (clear error instead of an opaque API failure), but intentionally not rewritten, since silently changing a user-specified ref felt riskier than a clear error.

secretspec/src/provider/akv.rs
phoem: Added ?suffix= as an explicit override; kept the dotted-hostname form working since it's already documented/tested behavior rather than removing it outright

secretspec/src/provider/akv.rs
phoem: Agreed, tracked as a follow-up — will need a ref.version → Key Vault version-suffix mapping.

secretspec/src/provider/akv.rs
phoem: consolidated into AuthMethod::as_str()/FromStr so the two directions can't drift

docs/src/content/docs/providers/akv.md
phoem: swapped for a relevant get example

secretspec/src/provider/akv.rs
phoem: Fixed the wording in reference/providers.md to say system-assigned, matching the provider page

secretspec/src/provider/akv.rs
phoem: Good callout — punting this to a follow-up rather than folding a caching layer into this PR

PR #134, opened 2026-07-14
feat(awssm): support KMS key and tags on secret creation

### LINKED ISSUE #133: awssm: support tags and KMS key when creating secrets
Many AWS orgs enforce tag-on-create: an SCP or IAM condition denies `secretsmanager:CreateSecret` unless required tags (`aws:RequestTag/...`) are present in the same call, and often a customer-managed KMS key is required too. In those accounts every `secretspec set` that creates a secret fails with `AccessDeniedException`, and there's currently no way to tag secretspec-created secrets or pick their key.

I have a working implementation that extends the provider URI, following the `?prefix=` pattern from #92:
```
awssm://prod@us-east-1?kms_key_id=alias/my-key&tag.team=platform&tag.env=prod
``` 

Tags and the key apply only on create (`PutSecretValue` accepts neither), and nothing changes when the params are absent. It's tested against a real org with a "tag-on-create" SCP: bare creates are denied, tagged ones go through.

Before opening the PR I wanted to check the shape with you. Unless I've missed something, I believe the alternative approach would be per-secret tags in `secretspec.toml`, but that touches the core schema and every SDK, and puts AWS-specific fields in the otherwise provider-agnostic manifest. Provider-level felt like the right first step since guardrail tags usually vary per environment rather than per secret, and per-secret overrides could still layer on later. Happy to to help with either, but wanted to see if this was even of interest and if it is what you would prefer.

domenkozar:
## Summary

Adds `kms_key_id` and `tag.NAME=VALUE` query parameters to the AWS Secrets Manager (`awssm`) provider:

```
awssm://prod@us-east-1?kms_key_id=alias/my-key&tag.team=platform&tag.env=prod
```

Fixes #133 

Both are applied **only when secretspec creates a secret** (`CreateSecret`). They must ride the `CreateSecret` call itself so accounts with "tag-on-create" guardrails — an SCP or IAM condition denying `CreateSecret` unless required `aws:RequestTag/*` tags (and often a customer-managed KMS key) are present in the same call — can store secrets. Updating a value (`PutSecretValue`) accepts neither, so a pre-existing secret keeps the key and tags it was created with.

## Details

- `tag.NAME=VALUE` pairs are collected from the query into a `BTreeMap`, so `uri()` reconstruction is deterministic for the audit log. Empty tag values (`tag.env=`) are preserved, since AWS accepts them.
- `uri()` now round-trips `prefix`, `kms_key_id`, and every tag.
- Docs updated: provider page URI format, examples, IAM notes (`TagResource`, `kms:GenerateDataKey`/`kms:Decrypt`), and a note that these pair well with a checked-in provider alias.

## Testing

- `cargo test --package secretspec provider::awssm` — 19 tests pass, including URI round-trip and empty-tag-value cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- The `awssm` provider accepts `kms_key_id` and `tag.NAME=VALUE` query
+  parameters (e.g. `awssm://prod@us-east-1?kms_key_id=alias/my-key&tag.team=platform`).
+  Both are applied only when secretspec creates a secret, so accounts that enforce
+  a customer-managed KMS key or "tag-on-create" guardrails (an SCP requiring
+  `aws:RequestTag/*` on `CreateSecret`) can now store secrets. A pre-existing
+  secret keeps the key and tags it was created with.
+
 ### Changed
 - A `ref` routed at a single store (an explicit `--provider`, a single-provider
   chain, or the default provider) is now checked up front, before any store is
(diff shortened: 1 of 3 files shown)


domenkozar:

PR #154, opened 2026-07-16
feat: add Infisical provider (infisical://)

### LINKED ISSUE #17: Support Infisical
[Infisical](https://infisical.com)

ojhermann:
## Summary

Adds an `infisical` provider for [Infisical](https://infisical.com) — Cloud and self-hosted — gated behind an `infisical` Cargo feature (in the default set alongside gcsm/awssm/vault/bws/akv).

It speaks Infisical's **v4** REST API directly over `reqwest`, which the vault provider already depends on, so the feature adds **no new dependency and leaves `Cargo.lock` untouched**. (The official `infisical` Rust SDK is 0.0.3, last released Sep 2025, and hardcodes the deprecated `/api/v3/secrets/raw` endpoints in all five secrets operations, so it isn't used.)

Closes #17

## URI format

```
infisical://[host]/{project-id}[?env=slug&path=/prefix&tls=false]
```

```bash
infisical://app.infisical.com/7e2f1a4c-...            # Infisical Cloud (US), the default host
infisical://eu.infisical.com/7e2f1a4c-...             # Infisical Cloud (EU)
infisical://vault.example.com/7e2f1a4c-...?env=prod   # read every profile from one environment
infisical://localhost:8080/7e2f1a4c-...?tls=false     # self-hosted over plain HTTP
```

The project is Infisical's project **UUID**; its v4 API takes no project slug (a slug in `projectId` 404s — verified). Without a host, the provider reads `INFISICAL_DOMAIN`, then defaults to `app.infisical.com`.

## Profiles map to environments

A SecretSpec profile names the Infisical environment, so a `production` profile reads the `production` environment — Infisical's environments are what profiles already mean. A project whose environments don't correspond to profiles pins one with `?env=`.

The profile names the **folder as well as** the environment, so pinning `?env=` cannot make two profiles resolve to the same secret. It looks redundant in the default layout and is load-bearing: without it, `?env=` makes `(project, "dev", KEY)` and `(project, "prod", KEY)` collide.

**Storage:** secret `{key}` in folder `{path}/{project}/{profile}` (default prefix `/secretspec`, overridable with `?path=`), in the environment named by the profile or `?env=`.

Keys are stored **verbatim** — Infisical constrains a key no further than being non-empty (lowercase, dots, `__`, spaces and Unicode all verified accepted), so nothing is rewritten and two keys cannot collide. Folder names are narrower (`[A-Za-z0-9_-]`), so a project or profile Infisical can't spell is **refused rather than rewritten**, which would let two projects share a folder.

## Authentication

Universal Auth (machine identity) via `client_id` + `client_secret`, exchanged once per run for a short-lived token; or a ready-made `token`, which is what CI usually has. Credentials come from the provider alias (so they can live in the keyring via `[providers]` `env`) or from `INFISICAL_CLIENT_ID` / `INFISICAL_CLIENT_SECRET` / `INFISICAL_TOKEN`. A half-specified identity errors naming the missing half rather than falling back to an anonymous request.

Service tokens are not supported: Infisical deprecated them in favour of machine identities.

## Refs

`ref = { item = "/infra/shared/DB_PASSWORD" }` names a secret by its own coordinates; `version` pins a version (read-only — a past version can't be rewritten). A leading slash names the environment's root, without one the folder is read under the configured prefix. Infisical values have no sub-components, so `field`, `section` and `vault` are rejected.

## Decisions worth flagging for review

- **Not-found is read from the status, never the message.** Infisical answers a missing secret, folder, environment and project with an indistinguishable bare 404, so all four are taken as "nothing stored here". A misspelled `?env=` therefore reads as an unset secret until a write reports it by name. Documented as a limitation; the alternative was matching error prose.
- **`expandSecretReferences=true`**, matching Infisical's own default and CLI — a `${...}` reference is a feature its users configure deliberately, so passing `false` would hand back a literal `${DB_USER}`.
- **A withheld value is refused, not passed on.** An identity that may see a secret exists but not read it still gets **HTTP 200**, carrying the literal string `<hidden-by-infisical>` with `secretValueHidden` set. Exporting that into the process `secretspec run` starts would be worse than failing, so it's reported as the refusal it is. (`viewSecretValue=true` alone does not prevent this.)
- **A write held for approval doesn't report success.** A project under an approval policy answers a write with 200 and an `approval` object, storing nothing until a human merges it.
- **`get_many` lists each folder once** rather than one round trip per secret — Infisical has no fetch-these-N-names endpoint, and its cloud rate limits are per-minute. Versioned refs fall back to the shared `get_each`, which already dedupes and parallelises.
- **Folders are created, not assumed.** Writing into a folder that doesn't exist 404s and stores nothing, so `set` creates the chain (one call creates every level) and lets the retry decide, rather than reading messages to tell "already exists" from a real fault.

## Testing

`cargo test --all --all-features`: **504 pass**, including provider integration tests run against **both** Infisical Cloud and self-hosted `v0.162.7`, which agreed on every behaviour tested.

```bash
export INFISICAL_CLIENT_ID=... INFISICAL_CLIENT_SECRET=...   # or INFISICAL_TOKEN
export INFISICAL_TEST_PROJECT=<project-uuid>
export INFISICAL_TEST_HOST=app.infisical.com                 # or localhost:8080 for self-hosted
export INFISICAL_TEST_ENV=dev                                # tests pin an env; profiles still separate by folder
SECRETSPEC_TEST_PROVIDERS=infisical cargo test --package secretspec --all-features
```

Also adds `test_all_providers_isolate_profiles` to the generic suite: one key written under three profiles, each asserted to keep its own value. It fails if a provider collapses profiles onto one secret — a class of bug that is invisible to a single-profile test. `dotenv`/`env` are named as flat stores rather than detected, since a store that collapsed its profiles by accident looks exactly like one that never had them.

## Known limitations

- The project is addressed by UUID; the v4 API accepts no slug
- A project or profile not spellable as an Infisical folder is refused rather than rewritten
- Refs need `?env=`, having no profile to name an environment with
- `secretspec import infisical://…` is unsupported: the provider doesn't enumerate secrets, so there's nothing for import to discover (`reflect()` follows the vault provider in not being implemented)
- `INFISICAL_DOMAIN` names a host, not a sub-path; a trailing `/api` is accepted, since Infisical's CLI takes the domain in that form
- **The withheld-value and approval-policy paths are implemented and unit-tested, but not verified against a live paid tier.** Custom roles are Pro-gated and approval workflows are Enterprise-gated; both guards are built against response shapes observed from the real API and the documented schema respectively. The restricted-role case ships as a *gated* integration test (`INFISICAL_TEST_NOREAD_CLIENT_ID`/`_SECRET`) rather than an absent one, so it runs wherever such a role exists.

### DIFF
--- a/docs/src/content/docs/providers/infisical.md
+++ b/docs/src/content/docs/providers/infisical.md
@@ -1,0 +1,211 @@
+---
+title: Infisical Provider
+description: Infisical integration
+---
+
+The Infisical provider integrates with [Infisical](https://infisical.com) over its REST API, for
+both Infisical Cloud and self-hosted instances.
+
+:::note[Version compatibility]
+The Infisical provider is an upcoming SecretSpec 0.16 feature and is not available
+in SecretSpec 0.15.
+:::
+
+## Prerequisites
+
+- An Infisical project
+- A machine identity with access to it (see [Authentication](#authentication))
+- Build with `--features infisical`
+
+## Configuration
+
+### URI Format
+
+```
+infisical://[host]/{project-id}[?env=slug&path=/prefix&tls=false]
+```
+
+- `host`: the Infisical instance (falls back to `INFISICAL_DOMAIN`, then `app.infisical.com`)
+- `{project-id}`: the project's **UUID**, from Project Settings → Project ID
+- `?env=`: environment slug. Without it, the SecretSpec profile names the environment
+- `?path=`: folder prefix holding SecretSpec's secrets (default: `/secretspec`)
+- `?tls=false`: disable TLS, for self-hosted instances served over plain HTTP
+
+Infisical's API addresses a project by UUID, not by the slug shown in its UI.
+
+### Examples
+
+```bash
+# Infisical Cloud (US, and the default host)
+secretspec set DATABASE_URL --provider infisical://app.infisical.com/7e2f1a4c-...
+
+# Infisical Cloud (EU)
+secretspec check --provider infisical://eu.infisical.com/7e2f1a4c-...
+
+# Self-hosted
+secretspec run --provider infisical://vault.example.com/7e2f1a4c-... -- npm start
+
+# Self-hosted over plain HTTP
+secretspec run --provider "infisical://localhost:8080/7e2f1a4c-...?tls=false" -- npm start
+```
+
+## Profiles and environments
+
+By default a SecretSpec profile names the Infisical environment: a `production` profile reads the
+`production` environment, and `dev` reads `dev`. New Infisical projects come with `dev`, `staging`
+and `prod`, so this works out of the box for profiles named after them.
+
+A project whose environments do not correspond to profiles pins one with `?env=`:
+
+```bash
+# Every profile reads Infisical's "dev" environment
+secretspec run --provider "infisical://app.infisical.com/7e2f1a4c-...?env=dev" -- npm start
+```
+
+Profiles stay separate either way: the profile names the folder as well as the environment, so
+pinning `?env=` cannot make two profiles share a secret.
+
+To route each profile to a different environment, give each one its own alias:
+
+```toml
+[providers]
+infisical_dev = "infisical://app.infisical.com/7e2f1a4c-...?env=dev"
+infisical_prod = "infisical://app.infisical.com/7e2f1a4c-...?env=prod"
+
+[profiles.production]
+DATABASE_URL = { description = "Production database", providers = ["infisical_prod"] }
+```
+
+### Secret Naming
+
+Secrets are stored under the folder `{path}/{project}/{profile}`, in the environment named by the
+profile (or by `?env=`):
+
+```
+project "myapp", profile "prod", key "DATABASE_URL"
+  -> environment prod
+     folder      /secretspec/myapp/prod
+     key         DATABASE_URL
+```
+
+Keys are stored exactly as written: Infisical accepts any non-empty key, so nothing is rewritten
+and two keys can never collide. Folder names are narrower — letters, digits, dashes and
+underscores — so a project or profile Infisical cannot spell is refused rather than quietly
+renamed.
+
+Folders are created as needed when writing a secret.
+
+## Secret References
+
+A secret can name one Infisical secret by its own coordinates, instead of SecretSpec's layout:
+
+```toml
+[profiles.production]
+DATABASE_URL = { description = "Postgres DSN", ref = { item = "/infra/shared/DB_PASSWORD" } }
+API_KEY = { description = "Pinned key", ref = { item = "/infra/API_KEY", version = "3" } }
+```
+
+- `item`: the folder and key. A leading slash names the environment's root — `/infra/shared/DB_PASSWORD`
+  is read from `/infra/shared`. Without one, the folder is read under the configured prefix, so
+  `team/DB_PASSWORD` means `/secretspec/team/DB_PASSWORD` and a bare `DB_PASSWORD` sits at
+  `/secretspec` itself
+- `version`: an Infisical secret version. Version-pinned refs are read-only, since a past version cannot be rewritten
+
+A ref has no profile to name an environment with, so the provider URI must pin one with `?env=`.
+Infisical secrets are single values with no sub-components, so `field`, `section` and `vault` are
+rejected.
+
+## Secret references inside values
+
+Values are read with Infisical's own `${...}` references expanded, matching its CLI, so a value of
+`postgres://${DB_USER}@host` arrives resolved.
+
+## Authentication
+
+### Universal Auth (machine identity)
+
+Create a machine identity in Infisical, grant it access to the project, and set:
+
+```bash
+export INFISICAL_CLIENT_ID=...
+export INFISICAL_CLIENT_SECRET=...
+```
+
+The provider exchanges these for a short-lived access token once per run.
+
+### Token
+
+A token minted elsewhere is used as it stands, which is what CI often has to hand:
+
+```bash
+export INFISICAL_TOKEN=...
+```
+
+Service tokens are not supported: Infisical deprecated them in favour of machine identities.
+
+### Sourcing credentials from another provider
+
+A machine identity's credentials can live in another store rather than in the environment, declared
+as [provider credentials](/concepts/providers/#provider-credentials):
+
+```toml title="secretspec.toml"
+[providers]
+infisical = { uri = "infisical://app.infisical.com/7e2f1a4c-...", credentials = {
+  client_id = "keyring",
+  client_secret = "keyring",
+} }
+```
+
+The provider declares three credentials: `client_id` and `client_secret` for Universal Auth, and
+`token` for a ready-made access token. Each falls back to its environment variable
+(`INFISICAL_CLIENT_ID`, `INFISICAL_CLIENT_SECRET`, `INFISICAL_TOKEN`) when not declared here.
+
+`secretspec config provider login infisical` prompts for each and stores it where resolution later
+reads it from.
+
+## Usage
+
+```bash
+# Store a secret
+secretspec set DATABASE_URL --provider infisical://app.infisical.com/7e2f1a4c-...
+
+# Verify every secret is set
+secretspec check --provider infisical://app.infisical.com/7e2f1a4c-...
+
+# Run with secrets injected
+secretspec run --provider infisical://app.infisical.com/7e2f1a4c-... -- npm start
+```
+
+Secrets sharing a folder are fetched in one request, so a run costs one call per folder rather than
+one per secret.
+
+## Self-hosting
+
+Point the URI at the instance, or set `INFISICAL_DOMAIN`:
+
+```bash
+export INFISICAL_DOMAIN=https://vault.example.com
+secretspec run --provider "infisical:///7e2f1a4c-..." -- npm start
+```
+
+## Approval policies
+
+A project under an approval policy turns a write into a change request: Infisical stores nothing
+until a human merges it. `secretspec set` reports that rather than claiming the secret was stored,
+so the value is written once the request is approved.
+
+## Limitations
+
+- The project is addressed by UUID; Infisical's API does not accept a project slug
+- A project or profile whose name is not spellable as an Infisical folder (letters, digits, dashes,
+  underscores) is refused rather than rewritten
+- Refs need `?env=`, having no profile to name an environment with
+- `secretspec import infisical://…` is not supported: the provider does not enumerate existing
+  secrets, so import has nothing to discover
+- `INFISICAL_DOMAIN` names a host, not a path: an instance served under a sub-path
+  (`https://example.com/infisical`) is not addressable. A trailing `/api` is the exception and is
+  accepted, since Infisical's own CLI takes the domain in that form
+- An environment or project that does not exist reads as an unset secret rather than an error:
+  Infisical answers a missing secret, folder, environment and project with the same bare 404.
+  Writing one reports the environment by name. If a profile does not match an environment slug —
+  Infisical's own default projects use `dev`, `staging` and `prod` — pin the right one with `?env=`
--- a/docs/src/content/docs/providers/infisical.md
+++ b/docs/src/content/docs/providers/infisical.md
@@ -25,7 +25,8 @@ in SecretSpec 0.15.
 infisical://[host]/{project-id}[?env=slug&path=/prefix&tls=false]
 ```
 
-- `host`: the Infisical instance (falls back to `INFISICAL_DOMAIN`, then `app.infisical.com`)
+- `host`: the Infisical instance (falls back to `INFISICAL_DOMAIN`, then the legacy
+  `INFISICAL_API_URL`, then `app.infisical.com`)
 - `{project-id}`: the project's **UUID**, from Project Settings → Project ID
 - `?env=`: environment slug. Without it, the SecretSpec profile names the environment
 - `?path=`: folder prefix holding SecretSpec's secrets (default: `/secretspec`)
@@ -115,6 +116,12 @@ A ref has no profile to name an environment with, so the provider URI must pin o
 Infisical secrets are single values with no sub-components, so `field`, `section` and `vault` are
 rejected.
 
+## Imported folders
+
+A folder that imports another resolves the imported keys too, with Infisical's own precedence: a
+secret defined directly in the folder wins over an imported one, and a later import wins over an
+earlier one. This matches their CLI, so a value reads the same way through either tool.
+
 ## Secret references inside values
 
 Values are read with Infisical's own `${...}` references expanded, matching its CLI, so a value of
@@ -188,6 +195,9 @@ export INFISICAL_DOMAIN=https://vault.example.com
 secretspec run --provider "infisical:///7e2f1a4c-..." -- npm start
 ```
 
+Infisical's legacy `INFISICAL_API_URL` is honoured too, so an instance already configured for
+their CLI works unchanged. `INFISICAL_DOMAIN` wins when both are set, matching the CLI.
+
 ## Approval policies
 
 A project under an approval policy turns a write into a change request: Infisical stores nothing
@@ -202,7 +212,7 @@ so the value is written once the request is approved.
 - Refs need `?env=`, having no profile to name an environment with
 - `secretspec import infisical://…` is not supported: the provider does not enumerate existing
   secrets, so import has nothing to discover
-- `INFISICAL_DOMAIN` names a host, not a path: an instance served under a sub-path
+- The domain variables name a host, not a path: an instance served under a sub-path
   (`https://example.com/infisical`) is not addressable. A trailing `/api` is the exception and is
   accepted, since Infisical's own CLI takes the domain in that form
 - An environment or project that does not exist reads as an unset secret rather than an error:
--- a/secretspec/src/provider/infisical.rs
+++ b/secretspec/src/provider/infisical.rs
@@ -31,8 +31,8 @@
 //! - `tls` -- enable TLS: `true` (default) or `false`, for self-hosted
 //!   instances served over plain HTTP.
 //!
-//! When no host is given, falls back to `INFISICAL_DOMAIN`, then to
-//! `app.infisical.com`.
+//! When no host is given, falls back to `INFISICAL_DOMAIN`, then to Infisical's
+//! legacy `INFISICAL_API_URL`, then to `app.infisical.com`.
 //!
 //! # Examples
 //!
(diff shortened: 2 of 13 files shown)


.claude/settings.local.json
```diff
@@ -3,8 +3,10 @@
     "allow": [
       "Bash(mkdir:*)",
       "Bash(jq:*)",
-      "Bash(cat:*)"
+      "Bash(cat:*)",
+      "Bash(git -C /home/otto/secretspec worktree prune)",
+      "Bash(git -C /home/otto/secretspec worktree add --detach /tmp/claude-1000/-home-otto-secretspec/b23d250b-f65c-431d-a55e-2cd240975739/scratchpad/wt c82f6a9)"
```
domenkozar: this should be gone

docs/src/content/docs/providers/infisical.md
```diff
@@ -0,0 +1,206 @@
...
+
+Service tokens are not supported: Infisical deprecated them in favour of machine identities.
+
+### Sourcing credentials from another provider
+
+A machine identity's credentials can live in another store rather than in the environment:
+
+```toml
+[providers]
+infisical = { uri = "infisical://app.infisical.com/7e2f1a4c-...", env = {
```
domenkozar: `env` is now called `credentials`

domenkozar (changes requested):

.claude/settings.local.json
ojhermann: Gone — reverted to the version on `main` and amended out of the commit, so it is not in the history. It was local editor config that should never have been staged; sorry for the noise.

docs/src/content/docs/providers/infisical.md
ojhermann: Fixed. That example was stale twice over — it used the old `env` key *and* env-var-style names. It now matches the `credentials` shape the other provider docs use:

```toml title="secretspec.toml"
[providers]
infisical = { uri = "infisical://app.infisical.com/7e2f1a4c-...", credentials = {
  client_id = "keyring",
  client_secret = "keyring",
} }
```

The provider registers `client_id`, `client_secret` and `token` (`infisical.rs`), so those are the names now documented, each falling back to its environment variable via `credential_or_env`. I also linked it to `/concepts/providers/#provider-credentials` to match the 1Password and BWS pages.

Only the docs were wrong — the provider code already used the post-rename API.

ojhermann:
Both review comments addressed, and rebased onto `main` to clear the conflict. 517 tests pass; `cargo fmt --check` clean.

**Review fixes**
- `.claude/settings.local.json` — reverted to `main`'s version and amended out of the commit entirely.
- The credentials example now uses `credentials = { client_id = ..., client_secret = ... }` instead of the old `env` key with env-var-style names, matching the other provider pages.

**Rebase — one thing worth a look**

The conflict was with `docs: showcase SecretSpec 0.15 on landing page`, which rewrote `index.astro`. I took your version of the page wholesale and re-applied the Infisical additions on top, rather than hand-merging the hunks. Two consequences:

- **Counts went 13 → 14.** I confirmed the number against the provider registry rather than the prose: `main` has 13 providers (`secretspec/src/provider/*.rs`, excluding `macros.rs`), so Infisical makes 14. Updated in all five places plus the bento big-num.
- **My branch's bento list had silently dropped Azure.** Your rewritten list is the correct 13 and includes it; taking your side fixed that as a side effect. It now reads as 14 with Infisical inserted.

Both `secretspec config init` blocks on the landing page were shortened by your rewrite and no longer enumerate providers, so there was nothing to add there.

I left the pre-existing `Gopass (0.15+)` label in `astro.config.ts` alone — now that 0.15 has shipped it presumably wants the durable "Available since" wording per `CLAUDE.md`, but that felt like separate cleanup rather than something to fold into this PR. Happy to include it if you'd rather it went in here.

The `(0.16+)` labels on Infisical stay, since it is still unreleased.

secretspec/src/provider/infisical.rs
```diff
@@ -0,0 +1,1296 @@
...
+        if scheme != "infisical" {
+            return Err(SecretSpecError::ProviderOperationFailed(format!(
+                "Invalid scheme '{scheme}' for infisical provider. Expected 'infisical'."
+            )));
+        }
+
+        // An unreadable `tls` is refused rather than read as one of its two
+        // meanings: the knob is explicit, so a typo in it should be too.
+        let use_tls = url
+            .query_pairs()
```
domenkozar, line 188: The standalone `infisical` feature currently does not compile. `ProviderUrl::query_pairs()` is gated to `awssm`/`vault`/tests, and `ProviderUrl::port()` is gated to `vault`/tests, so `cargo check --package secretspec --no-default-features --features infisical` fails here and at the `port()` call. Please add `feature = "infisical"` to both accessor cfgs (and ideally add a non-test feature check, since `cfg(test)` masks this).

secretspec/src/provider/infisical.rs
```diff
@@ -0,0 +1,1296 @@
...
+        let body = response.text().await.unwrap_or_default();
+
+        match status {
+            StatusCode::OK => {
+                let parsed: serde_json::Value = serde_json::from_str(&body).map_err(|e| {
+                    SecretSpecError::ProviderOperationFailed(format!(
+                        "Failed to parse Infisical response: {e}"
+                    ))
+                })?;
+                let secrets = parsed["secrets"].as_array().ok_or_else(|| {
```
domenkozar, line 651: This drops imported Infisical secrets during batch resolution. The v4 list response returns direct values in `secrets` and imported values separately in `imports` ([API response](https://infisical.com/docs/api-reference/endpoints/secrets/list)); `get_many()` uses only this method, so `check`/`run` can report an imported secret missing even though single-secret retrieval finds it. Please merge the imported secret entries into `listed` as well, preserving Infisical's precedence rules.

secretspec/src/provider/infisical.rs
```diff
@@ -0,0 +1,1296 @@
...
+        if let Some(token) = self.token.get() {
+            return Ok(token);
+        }
+        // `credential_or_env` never yields an empty value, so a blank
+        // INFISICAL_TOKEN reads as absent rather than as a broken token.
+        let token = match credential_or_env(&self.credentials, TOKEN, INFISICAL_TOKEN_ENV) {
+            Some(token) => SecretString::new(token.into()),
+            None => self.login().await?,
+        };
+        // A concurrent login would have produced an equivalent token.
```
domenkozar: `OnceLock` does not actually serialize login here: concurrent callers all pass `get()`, each awaits `login()`, and only afterward race in `get_or_init`. `get_many()` deliberately fetches distinct versioned refs concurrently, so Universal Auth can be exchanged multiple times. This can fail outright when the client secret has a one-use limit, which Infisical supports and recommends for some setups ([Universal Auth guide](https://infisical.com/docs/documentation/platform/identities/universal-auth)). Please synchronize the async initialization itself rather than only storing the completed token.

secretspec/src/provider/infisical.rs
```diff
@@ -0,0 +1,1296 @@
...
+/// names one.
+const DEFAULT_HOST: &str = "app.infisical.com";
+
+const CLIENT_ID: &str = "client_id";
+const CLIENT_SECRET: &str = "client_secret";
+const TOKEN: &str = "token";
+const INFISICAL_CLIENT_ID_ENV: &str = "INFISICAL_CLIENT_ID";
+const INFISICAL_CLIENT_SECRET_ENV: &str = "INFISICAL_CLIENT_SECRET";
+const INFISICAL_TOKEN_ENV: &str = "INFISICAL_TOKEN";
+const INFISICAL_DOMAIN_ENV: &str = "INFISICAL_DOMAIN";
```
domenkozar: Infisical's documented global domain variable is `INFISICAL_API_URL`, not `INFISICAL_DOMAIN` ([CLI domain configuration](https://infisical.com/docs/cli/usage)). As written, existing EU/self-hosted setups using the standard variable silently fall back to US Cloud. Please recognize `INFISICAL_API_URL`; `INFISICAL_DOMAIN` could remain as a compatibility alias if desired.

secretspec/src/provider/infisical.rs
ojhermann: Confirmed and fixed — I reproduced your exact command and got E0599 on both. Added `feature = "infisical"` to the `port()` and `query_pairs()` cfgs.

You were right about the cause, and it is the interesting part: `cfg(test)` meant my 517 tests could never see this. The suite builds with every provider on *and* `cfg(test)` set, so a gate of `any(feature = "...", test)` is satisfied twice over — the one build configuration that breaks is the one nobody runs.

So I took the "ideally" seriously and added a `features` job to `test.yml` that runs `cargo check --no-default-features --features X` for each of the eight. I checked the other seven first rather than assume: `cli`, `keyring`, `gcsm`, `awssm`, `vault`, `bws` and `akv` all already pass, so the job goes green on this PR rather than dragging in unrelated breakage. It is a `cargo check`, not a build, so it should be cheap.

Happy to drop the CI job into its own PR if you would rather keep this one to the provider.

secretspec/src/provider/infisical.rs
ojhermann: Confirmed and fixed. You are right that `get_many()` is the only caller of `list_async`, so `check`/`run` would call an imported secret missing while `get` returned it.

Worth noting how bad it was: `includeImports` **defaults to `true`**, so the imported values were already in the response I parsed — I read `parsed["secrets"]` and dropped `parsed["imports"]` on the floor. Nothing was being under-requested; it was being discarded.

For precedence I mirrored the CLI's `InjectRawImportedSecret` rather than inventing a rule, since the provider claims elsewhere to resolve values the way their tool does:

```go
hasOverriden := make(map[string]bool)
for _, sec := range secrets { hasOverriden[sec.Key] = true }   // direct wins
for i := len(importedSecrets) - 1; i >= 0; i-- {               // reverse: later import wins
```

So: a direct secret beats any import, and a later import beats an earlier one. Withheld imported values are refused exactly as direct ones are — an imported key the identity may not read must not arrive as the literal `<hidden-by-infisical>`.

Five tests cover it (merged, direct-wins, later-import-wins, withheld-refused, no-`imports`-key). I mutation-tested the two precedence rules: dropping the `.rev()` fails exactly the later-import test, and removing the direct-wins guard fails both. Also documented on the provider page and in the module docs.

secretspec/src/provider/infisical.rs
ojhermann: Confirmed and fixed. Your reading of the flow is exactly right: every caller passes `get()`, each awaits its own `login()`, and they only race at `get_or_init` — by which point the surplus token exchanges have already happened. `get_each` spawns a thread per address group, so this is the normal path for a batch read, not a corner case.

Now a `tokio::sync::OnceCell`, whose `get_or_try_init` serializes the initializer itself rather than the result:

```rust
self.token
    .get_or_try_init(|| async {
        match credential_or_env(&self.credentials, TOKEN, INFISICAL_TOKEN_ENV) {
            Some(token) => Ok(SecretString::new(token.into())),
            None => self.login().await,
        }
    })
    .await
```

A concurrent caller now waits for the in-flight exchange and takes its token. On failure the cell stays empty, so a later call retries rather than caching the error. It also deletes the manual get/check/init dance.

`tokio/sync` was already on transitively via reqwest, but I added it explicitly to the `infisical` feature — inheriting a feature from a sibling dependency is the same shape of latent breakage as the missing gate above. `Cargo.lock` is unchanged.

One thing I did **not** do: add a test. Counting logins needs an HTTP mock, and there is no mocking dependency in the tree today (`vault.rs` has no test module at all, so this seems to be the norm for the reqwest-based providers). Adding `wiremock` for a single test felt like the wrong trade against this PR adding no dependencies — the fix is structural rather than behavioural. Say the word if you would prefer the dependency and the test.

secretspec/src/provider/infisical.rs
ojhermann: Fixed — but I want to flag that the premise is inverted, because it affects which name is primary.

`INFISICAL_DOMAIN` is Infisical's **current** variable; `INFISICAL_API_URL` is the legacy one. From their CLI source (`Infisical/cli`, `packages/util/constants.go`):

```go
INFISICAL_DOMAIN_ENV_NAME         = "INFISICAL_DOMAIN"
LEGACY_INFISICAL_API_URL_ENV_NAME = "INFISICAL_API_URL" // superseded by INFISICAL_DOMAIN; kept for backwards compatibility
```

and `packages/util/helper.go`:

```go
// DomainEnvNames lists the env vars that configure the Infisical domain, in
// precedence order: INFISICAL_DOMAIN first, then the legacy INFISICAL_API_URL.
var DomainEnvNames = []string{INFISICAL_DOMAIN_ENV_NAME, LEGACY_INFISICAL_API_URL_ENV_NAME}
```

Their `--domain` flag help says the same: "Can also set via INFISICAL_DOMAIN environment variable".

**But your actual concern is real and I have fixed it**, just the other way round: reading only one name means a setup configured with the *other* silently lands on US Cloud. That is a bad failure either way, so the provider now reads both, in the CLI's precedence — `INFISICAL_DOMAIN`, then `INFISICAL_API_URL` — which means an instance already configured for their tool works unchanged whichever name it uses.

Also: an invalid domain now names the variable that actually set it, rather than always blaming `INFISICAL_DOMAIN`. The lookup is split into a `pick_domain(lookup)` taking a closure so precedence, trimming and blank-fallthrough are testable without mutating the process environment — four tests cover it. Docs updated.

If the docs page you were reading says otherwise, it may be worth an upstream report on their side; I went to the CLI source because their docs and their code have disagreed before in this area (the `/api` suffix handling was the last one).

ojhermann:
All five addressed in `e8564d0`. 527 tests pass (was 517), `cargo fmt --check` clean, `Cargo.lock` still untouched.

Four were straight hits — the two correctness ones (dropped imports, the login race) are exactly as you described, and I have replied inline on each with what changed. Two things worth pulling up to the top:

**The landing-page comment doesn't have an inline thread any more** — reverting the file removed the line it anchored to, so GitHub dropped it. I took your first option: the unversioned copy stays at **13**. I went looking for precedent and found `66bbab1` (gopass, while unreleased), where `index.astro` got *exactly one line* — the `providerMetadata` entry, no count bump, no prose edits. My diff to that file is now precisely that:

```diff
+  { slug: 'infisical', label: 'Infisical (0.16+)' },
```

How it happened, since it's the useful part: this was fallout from rebasing onto your 0.15 landing-page rewrite. I took your page and re-applied Infisical on top, then bumped 13→14 because I'd verified 13 against the provider registry. I checked the count was arithmetically right and never asked whether it was *true of the released version* — which is the only thing that rule cares about. The registry was the wrong oracle.

**The `INFISICAL_API_URL` premise is inverted**, and I've laid out the source inline. Short version: `INFISICAL_DOMAIN` is Infisical's current variable and `INFISICAL_API_URL` is the legacy one their CLI comments literally mark `// superseded by INFISICAL_DOMAIN; kept for backwards compatibility`. Your underlying concern is real either way, so the provider now reads **both**, in their CLI's precedence order. The outcome you asked for, with the primary/alias roles the other way round.

**On the feature gate — the part I'd want a second opinion on.** You called it: `cfg(test)` meant 517 tests could never see that break, because the suite builds with every provider on *and* `cfg(test)` set, so an `any(feature = "...", test)` gate is satisfied twice over. The one configuration that breaks is the one nobody runs. I added a `features` job to `test.yml` checking each of the eight standalone, and verified the other seven already pass so it goes green here rather than dragging in unrelated breakage. **Happy to split that into its own PR** if you'd rather this one stayed provider-only — it fixes a repo-wide blind spot, not an Infisical one.

**One thing I deliberately did not do:** add a test for the login race. Counting exchanges needs an HTTP mock, and there's no mocking dependency in the tree (`vault.rs` has no test module at all, so pure unit tests look like the norm for reqwest-based providers). Adding `wiremock` for one test seemed a poor trade against this PR adding no dependencies, and the fix is structural rather than behavioural — `get_or_try_init` serializes by construction. Say the word if you'd prefer the dependency and the test.

Everything else is covered by tests: five for import merging (mutation-tested — dropping the `.rev()` fails exactly the later-import test; removing the direct-wins guard fails both precedence tests), four for domain precedence.

ojhermann:
Two CI failures — one mine, one not.

**`features` — mine, fixed in `8718988`.** The job I added runs on a bare runner, where `keyring` cannot build: it links libdbus for its secret-service transport on Linux, and only the devenv job gets that from its shell. Now installed the way `go-embed.yml` and `ffi-build.yml` already do it:

```yaml
- name: Install system dependencies
  run: sudo apt-get update && sudo apt-get install -y libdbus-1-dev pkg-config
```

I reproduced the failure locally by dropping dbus off `PKG_CONFIG_PATH` (6 `dbus-1` errors) and confirmed it goes green with it. Worth admitting the shape of this one: the job passed for me because the nix shell I developed it in supplied dbus — my environment hid the missing dependency, which is precisely the mistake the job exists to catch. It has now caught it once before merging, which is something.

**`claude-review` — not mine, and I have deliberately not touched it.** It fails with:

> Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?

`claude-code-review.yml` does request `id-token: write`, so the permissions block is not the problem. GitHub does not issue OIDC tokens to workflows triggered by `pull_request` **from a fork**, whatever the workflow asks for — fork runs get a read-only token and no secrets. So this fails on any outside contributor's PR, not just this one.

The obvious fix — switching to `pull_request_target` — would run the workflow with repo secrets against untrusted fork code, so I am not going to send that as a drive-by. Flagging it as yours to decide.

`build`, `windows`, and `tests (ubuntu-latest)` all pass.

nickdichev:
I'm migrating from BWS to Infisical and I find the folder namespacing implemented by the provider to be a strange default. I understand the value there, but it wasn't immediately clear to me why my secrets in the root of my infisicial project weren't getting resolved.

The ability to import secrets in infisical was a good workaround though. I was able to create the expected folder structure and point at my project root.

Anyway, I'm curious if we should allow the provider config to pass values like `?layout=flat` to opt into a non-namespaced folder layout in infisical.

ojhermann:
Glad the imports got you unblocked @nickdichev. There's a more direct escape hatch you might find useful — a per-secret `ref` with a leading-slash `item` reads the environment root without any folder scaffolding:

```toml
DATABASE_URL = { ref = { item = "/DATABASE_URL" }, providers = ["infisical"] }
```

(with `?env=` pinned on the provider). It's per-secret rather than a whole-project switch, but there's no import tree to maintain.

On `?layout=flat`: I think a flat opt-in is reasonable. The namespacing default is there so multiple projects/profiles can't collide, and so pinning `?env=` still keeps profiles apart — but that's exactly what a flat, single-project store like a BWS migration doesn't need. @domenkozar, would you be open to it? Happy to put up a PR. One wrinkle worth noting: flat + a pinned `?env=` would collapse every profile onto the same key, since today the profile folder is what separates them — so it'd be an explicit "you're giving up profile separation" mode.

domenkozar:
Sure let's explore it.

ojhermann:
Opened #181 to track the `?layout=flat` investigation and a resolving PR. Design points and the profile-separation wrinkle are captured there — continuing in that thread.

PR #155, opened 2026-07-16
docs(ffi): document the request `mode` field, and test it

ojhermann:
`secretspec-ffi` is the only place in the repo that doesn't know its own request contract has a `mode` field. Both the module docs and the shipped C header document the request as exactly five fields:

```c
 * Request JSON (all fields optional):
 *   { "path": ".../secretspec.toml", "provider": "keyring://",
 *     "profile": "production", "reason": "boot", "no_values": false }
```

But `secretspec_resolve` hands its request straight to `secretspec::resolve_json`, which has accepted `mode` since `report()` landed — and every SDK already relies on it, including the four that reach it through this very ABI (`secretspec-hs`, `-go`, `-rb`, `-php`):

```haskell
-- secretspec-hs/src/SecretSpec.hs
report b = do
  resp <- callNative (requestBytes b (Just "report"))
```

The header isn't incidental — `stage-staticlib.sh` copies it into the Go SDK's `include/`, so it's the artifact a native consumer actually reads. Someone binding this ABI from C has no way to discover the report mode at all.

`tests/c_abi.rs` had the same blind spot: it covered `resolve`, `no_values`, missing-required, bad JSON and a missing manifest, but never `report`. So the one documented-contract gap was also the one untested path. This adds both halves.

### The part worth naming

`no_values` looks like it does the same job as `mode: "report"`. It doesn't, and the difference bites exactly when it matters:

| | `no_values: true` | `mode: "report"` |
|---|---|---|
| Shape | `ResolveResponse`, values blanked | `ResolutionReport` |
| `secrets` | object keyed by name | **array** of entries |
| `required` per secret | absent | **present** |
| `status` per secret | absent | `resolved` / `missing_required` / `missing_optional` |
| When a required secret is missing | `secrets` is **empty** | full inventory retained |

A preflight consumer reaching for `no_values` gets a response that goes empty precisely in the situation it was written to describe, and can never see requiredness.

### Tests

Three added to `tests/c_abi.rs`, all through the real `extern "C"` entry points:

- `report_mode_returns_requiredness_and_status` — pins the array shape, `required`, `status`, provenance, and that no value is carried.
- `report_mode_keeps_the_inventory_when_a_required_secret_is_missing` — pins the contrast with `resolve_missing_required_is_ok_envelope_with_error_list`, which asserts the opposite for `resolve`.
- `unknown_mode_yields_error_envelope` — pins that an unknown `mode` is `invalid_request` rather than a silent fallback to `resolve`.

I mutation-tested the two report tests by dropping the `mode` field: both fail, so they depend on report mode rather than passing incidentally.

Locally: 9/9 in `secretspec-ffi` (was 6), `cargo fmt --check` clean, `cargo clippy` clean for this crate, `RUSTDOCFLAGS="-D warnings" cargo doc` clean for the new intra-doc links.

### One judgement call for you

**No CHANGELOG entry.** `CLAUDE.md` says Rust commits get one, but also says to keep entries user-facing and leave out development and testing notes — and nothing here changes behaviour, so there's no user-facing change to report (`mode` already worked). Happy to add one if you'd rather; I'd rather ask than guess.

Documentation and tests only; no behaviour change.

### DIFF
--- a/secretspec-ffi/include/secretspec.h
+++ b/secretspec-ffi/include/secretspec.h
@@ -7,14 +7,37 @@
  *
  * Request JSON (all fields optional):
  *   { "path": ".../secretspec.toml", "provider": "keyring://",
- *     "profile": "production", "reason": "boot", "no_values": false }
+ *     "profile": "production", "reason": "boot", "no_values": false,
+ *     "mode": "resolve" }
+ *
+ * "mode" selects the response shape and defaults to "resolve":
+ *
+ *   "resolve"  the value-carrying resolve response. Set "no_values" to strip
+ *              the values from it.
+ *   "report"   the value-free resolution report: the inventory/preflight view
+ *              the CLI exposes as `check --json`.
+ *
+ * Any other value is rejected with an "invalid_request" error.
+ *
+ * "no_values" is NOT the same as "mode": "report". A "no_values" resolve blanks
+ * the values but keeps the resolve shape: its "secrets" is an object keyed by
+ * name, it never says whether a secret is *declared* required, and when a
+ * required secret is missing that object is empty. A report's "secrets" is an
+ * ARRAY of per-secret entries carrying "name", "required", "status"
+ * ("resolved" / "missing_required" / "missing_optional") and provenance, and
+ * lists every declared secret whether or not it resolved. "required" is
+ * reachable only via "report".
  *
  * Response envelope:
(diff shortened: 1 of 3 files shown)


domenkozar:
Thank you!

ojhermann:
> Thank you!

Happy to help; I love this project.

PR #157, opened 2026-07-17
test: property tests for the naming and encoding invariants

### LINKED ISSUE #156: Property-based tests for the injectivity and round-trip invariants the provider layer already states
## The gap

Two invariants hold the provider layer together, and both are stated in prose,
claimed over an infinite domain, and guarded by a handful of hand-picked
examples.

**Injectivity** — distinct `(project, profile, key)` triples must never collide
on one stored name. `akv.rs` states it directly:

> Encodes a component into a lowercase, Azure-compatible and **injective**
> representation.

**Inverse** — `uri()` must render what the parser read. Sixteen providers
implement it, and the intent is written down all over:

- `infisical.rs` — *"The inverse of parsing: every knob the config holds is rendered"*
- `akv.rs` — *"The inverse of `AuthMethod::as_str`"*
- `config.rs` — *"The two forms round-trip losslessly"*
- `provider/mod.rs`, on `QUERY_ENCODE_SET` — *"makes `ProviderUrl::encode_query`
  a true inverse of that parsing, so query values round-trip"*

Both are universally quantified claims. Nothing in the test suite quantifies
over anything: there is no `proptest`, `quickcheck`, `arbitrary` or `loom` in the
tree.

## Why I think it's worth your time

Not on principle — on this repo's own record. Consider what guards injectivity
today, after `Fix Azure Key Vault name collisions` rewrote ~150 lines of `akv.rs`
and reworked its docs:

```rust
#[test]
fn test_format_secret_name_prevents_boundary_delimiter_collision() {
    let trailing = AkvProvider::format_secret_name("a", "b-", "C").unwrap();
    let leading = AkvProvider::format_secret_name("a", "b", "_C").unwrap();
    assert_ne!(trailing, leading);
}

#[test]
fn test_format_secret_name_encodes_internal_delimiters() {
    let left = AkvProvider::format_secret_name("a", "b__c", "d").unwrap();
    ...
```

That second one is the counterexample from the review of #132, frozen into a
regression test:

> **Correctness — name mapping is not injective.** `_`→`-` sanitization plus the
> `--` delimiter means distinct triples collide: `format_secret_name("a","b__c","d")`
> and `format_secret_name("a","b","c__d")` both produce `secretspec--a--b--c--d`

That is the shape of the whole argument. A human read the code, constructed two
inputs colliding on one output, and the fix cost a 150-line rewrite of the naming
scheme. The tests now guarding the rewrite are `assert_ne!` over **one pair
each** — they encode the counterexamples that were already found, not the
property that was violated.

`proptest` finds that class automatically and shrinks to a minimal case. The
`"b__c"` collision is exactly what a generator over
`(project, profile, key)` triples produces in milliseconds.

The same pattern shows up on the inverse side. `QUERY_ENCODE_SET` exists because
query values were not round-tripping; it landed with the pass `store_dir`
feature, and its test loops over eight hand-chosen strings:

```rust
for store_dir in ["/custom/store", "/srv/store+1", "/data/a&b", "/data/a#b",
                  "/has space", "/pct%20literal", "/q?x=y", "/all/&+#%=? mix"] {
```

That list is a good list. It is also someone's guess at which characters matter,
maintained by hand, per provider, forever — and it only exists for `pass`.

## Proposal

Narrow, and reviewable in one sitting. I would start with the one that already
has a maintainer-authored counterexample in the record:

**1. `akv::format_secret_name` is injective.**
```rust
proptest! {
    #[test]
    fn distinct_triples_never_collide(a in triple(), b in triple()) {
        prop_assume!(a != b);
        prop_assert_ne!(format_secret_name(&a.0, &a.1, &a.2)?,
                        format_secret_name(&b.0, &b.1, &b.2)?);
    }
}
```

**2. `encode_query` / `query_pairs` are inverse** over arbitrary strings — the
invariant `QUERY_ENCODE_SET` already claims in prose, currently exercised only
via one provider's eight-string list.

**3. `parse(uri(cfg)) == parse(spec)`** for provider URI specs, generalizing the
per-provider round-trip tests. This is the biggest lift (it needs generators per
config) and the one I would do last, or drop.

Happy to land these one PR at a time, or just #1 as a proof and stop there if it
does not earn its keep.

## Costs, honestly

- **A new dev-dependency.** `proptest` (or `quickcheck`) in a repo that has none
  beyond std and `trybuild`. That is the real ask here, and a fair place to say
  no.
- **The generators are the work**; the properties are one-liners. Most of the
  diff would be `Arbitrary`/strategy code for config types.
- **CI time and flake risk** if generation is unbounded. Mitigable with a low
  case count and fixed seeds, but it is not free.
- **It buys nothing outside this family.** It would not catch a wrong API field,
  a missing feature gate, or a concurrency bug. It is aimed squarely at the
  inverse/injectivity invariants above — which happen to be the ones with a
  repeat failure history here.

If the dependency is unwelcome, there is a smaller version: the injectivity
property can be a plain unit test looping a few hundred generated triples with no
new crates. Less good, but it would have caught the `__` collision.

## Offer

Happy to implement whichever of the above you think is worth having, starting
with #1, if you would like it. Equally happy to hear this is not a direction you
want — it is a dependency and a style, and that is your call, so I would rather
ask before writing code than send an unsolicited PR.

ojhermann:
Closes #156.

Adds `proptest` as a dev-dependency and quantifies the two invariants from the issue: akv name injectivity, and the `encode_query` inverse. Seven properties, 0.59s.

I checked both against the **real pre-fix implementations** rather than trusting that they work. That turned out to matter — see the second section.

## What they catch

**akv name injectivity.** Restoring the old `_`→`-` mapping in `encode_name_component`:

```
Test failed: name "secretspec--A--A--a" did not decode back to its triple
minimal failing input: (project, profile, key) = ("A", "A", "a")
```

**Query encoding.** Restoring the plain URI encoder that `QUERY_ENCODE_SET` replaced:

```
Test failed: value "#" did not survive the round-trip through "keyring://?v=#"
minimal failing input: value = "#"
```

Both fail on essentially the first case and shrink to something minimal.

## The design bit worth reviewing

**Injectivity stated over pairs is the honest phrasing, but a bad test.** My first attempt was the obvious one:

```rust
fn distinct_triples_never_collide(a in triple(), b in triple()) {
    prop_assume!(a != b);
    prop_assert_ne!(format_secret_name(&a.0, ..)?, format_secret_name(&b.0, ..)?);
}
```

Run against the old `_`→`-` scheme, **it passes.** Two independently generated triples essentially never collide by chance — finding `("a","b__c","d")` vs `("a","b","c__d")` that way is a needle in a haystack. It looks like a test of injectivity and is close to vacuous.

So the load-bearing property here is a **left inverse** instead: a name decodes back to exactly the triple that built it. If every name decodes to one triple, no two triples can share a name — same guarantee, but every single generated case exercises it, and it's linear rather than quadratic to sample.

I kept the pairwise form too, because it states the guarantee in the shape the module actually claims ("injective") and would catch a collision introduced somewhere other than the encoding. But it only earns its place with a generator biased toward the `_`/`-` region where every historical collision has lived:

```rust
prop_oneof!["[A-Za-z0-9_-]{1,8}", "[_-]{1,3}"]
```

So biased, against the old scheme it finds one immediately:

```
distinct triples ("_", "-", "---") and ("_", "_-_", "-")
  both produced "secretspec-----------"
```

That bias is commented in the source, since it looks arbitrary otherwise and is the difference between a real test and a decorative one.

## Cost

- **8 new lockfile entries**, all dev-only: `proptest`, `bit-set`, `bit-vec`, `quick-error`, `rand_xorshift`, `rusty-fork`, `unarray`, `wait-timeout`. Nothing reaches anyone depending on `secretspec`.
- **0.59s** for the seven properties. Default 256 cases; happy to lower it.
- No `proptest-regressions` files yet. If you'd rather those weren't committed when a failure is found, that's a one-line `.gitignore` — tell me which you prefer.

## Not included

The third item from #156 — `parse(uri(cfg)) == parse(spec)` across every provider — isn't here. The trait object only exposes `uri()`, so what's reachable generically is idempotence (`uri(parse(uri(parse(s)))) == uri(parse(s))`), which is strictly weaker: it wouldn't catch a *dropped* field, because the loss happens in the first render and is stable after. Doing it properly needs per-provider config generators and `PartialEq` on the config types. Worth its own PR if you want it; I didn't want to smuggle that scope in here.

No CHANGELOG entry: test-only, no behaviour change, and the CHANGELOG asks for user-facing entries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -704,7 +704,16 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
 dependencies = [
- "bit-vec",
+ "bit-vec 0.6.3",
+]
+
+[[package]]
+name = "bit-set"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
+dependencies = [
+ "bit-vec 0.8.0",
 ]
 
 [[package]]
@@ -712,6 +721,12 @@ name = "bit-vec"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
+
+[[package]]
+name = "bit-vec"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
 name = "bitflags"
@@ -2001,7 +2016,7 @@ version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "531e46835a22af56d1e3b66f04844bed63158bc094a628bec1d321d9b4c44bf2"
 dependencies = [
- "bit-set",
+ "bit-set 0.5.3",
  "regex-automata",
  "regex-syntax",
 ]
@@ -3837,6 +3852,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "proptest"
+version = "1.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b45fcc2344c680f5025fe57779faef368840d0bd1f42f216291f0dc4ace4744"
+dependencies = [
+ "bit-set 0.8.0",
+ "bit-vec 0.8.0",
+ "bitflags",
+ "num-traits",
+ "rand 0.9.2",
+ "rand_chacha 0.9.0",
+ "rand_xorshift",
+ "regex-syntax",
+ "rusty-fork",
+ "tempfile",
+ "unarray",
+]
+
+[[package]]
 name = "pulldown-cmark"
 version = "0.9.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3903,6 +3937,12 @@ dependencies = [
  "quote",
  "syn 2.0.117",
 ]
+
+[[package]]
+name = "quick-error"
+version = "1.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quinn"
@@ -4056,6 +4096,15 @@ name = "rand_core"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c8d0fd677905edcbeedbf2edb6494d676f0e98d54d5cf9bda0b061cb8fb8aba"
+
+[[package]]
+name = "rand_xorshift"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "513962919efc330f829edb2535844d1b912b0fbe2ca165d613e4e8788bb05a5a"
+dependencies = [
+ "rand_core 0.9.5",
+]
 
 [[package]]
 name = "rayon"
@@ -4430,6 +4479,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
+name = "rusty-fork"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc6bf79ff24e648f6da1f8d1f011e9cac26491b619e6b9280f2b47f1774e6ee2"
+dependencies = [
+ "fnv",
+ "quick-error",
+ "tempfile",
+ "wait-timeout",
+]
+
+[[package]]
 name = "ryu"
 version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ojhermann:
Force-pushed to drop two `proptest-regressions/` files I had committed by accident (`git add -A` swept them up). They were seeds from failures of code I had *temporarily broken* to check the properties actually bite — not failures of anything in this PR, so they would have read as "what went wrong here?" for no reason.

Removing them also makes the note at the bottom of the description true again: there are no regression files, and the choice of whether to commit them or `.gitignore` them when a real failure turns up is still yours.

Now 5 files, +268/−2. Properties unchanged and still passing.

domenkozar:
Could we somehow write the property test once and run it through all registered providers?

ojhermann:
> Could we somehow write the property test once and run it through all registered providers?

I'll look into this

ojhermann:
@domenkozar — yes, prototyped it. One property test now enumerates `PROVIDER_REGISTRY`, constructs each provider offline (URL parsing only, no client/network), and asserts the naming invariants across all 14 registered providers in a single body.

What made it shareable: two small, pure trait methods —
- `is_valid_native_name(item)` — what names a store accepts (default: any non-empty; akv overrides with its letters/digits/hyphens rule).
- `convention_collapses_scope()` — whether a store drops project/profile (dotenv, env, bws).

With those, one test covers **determinism**, **charset-legality** (the invariant akv's Base32 scheme exists to satisfy — now free for every current and future provider), and **injectivity** (skipping the scope-collapsing stores). akv keeps its own Base32 round-trip proptest, since decodability needs knowledge of its scheme and doesn't generalize.

It also surfaced a real thing: **bws collapses project/profile like dotenv/env** but wasn't classified as flat (`FLAT_PROVIDERS` listed only those two). Worth a look regardless of the test question.

Your call on scope:
1. **Take the generic test** — the two trait methods + the shared proptest; one test across all providers for determinism/charset/injectivity, akv round-trip stays local. Prototyped and green; I can push it here.
2. **Keep naming tests provider-local** — the shared *encoding* test already covers the one truly shared function (`encode_query`); leave naming per-provider.

I lean 1 — charset coverage for every future provider is the real win — but the two methods are an API addition, so your call.

PR #163, opened 2026-07-17
feat: secret scopes (--scope), membership-only subsets of a profile

### LINKED ISSUE #137: Per-component / per-service secret scoping orthogonal to profiles
Mentioned in https://github.com/cachix/secretspec/issues/49#issuecomment-4559848026.

Benefits:
* Least privilege: each service could resolve and receive only the secrets it declares, instead of every process in the environment seeing every secret in the profile
* `secretspec check` could be modified so that developers touching only one service do not have to get blocked or prompted for other services' required secrets they don't have and shouldn't need
* Cleaner declaration: assuming N services and M actual profiles, if you were to do least-privilege manually with secretspec profiles, you would author N * M units instead of N + M units

Open questions:
* The motivating case is a single shell running several processes that each need a different subset, so more than one scope is live at once. Does secretspec resolve once and expose the union, leaving per-process routing to devenv, or does one invocation resolve and hand out several named scopes? Does `declare_secrets!` emit one struct, one per component, or a component-selected view?
* Should this use per-secret tags (e.g. `API_KEY = { ..., components = ["api", "worker"] }`) or `[components.api]` tables that list their own secrets?
* Does a component only filter membership, or also carry values, i.e. allowing you to override `default` / `required` / `providers` per service?

https://github.com/cachix/secretspec/issues/95 is tangentially related

ojhermann:
Implements the membership-only **scope** design from #137, following @domenkozar's [detailed steer](https://github.com/cachix/secretspec/pull/163#issuecomment-5006770221). Closes #137.

## What this adds

A `[scopes]` table names membership-only subsets of a profile's secrets, so a single service resolves only what it declares instead of the whole profile:

```toml
[profiles.default]
DATABASE_URL = { description = "Database", required = true }
API_KEY      = { description = "API key", required = true }
QUEUE_TOKEN  = { description = "Queue token", required = true }

[scopes.api]
secrets = ["DATABASE_URL", "API_KEY"]

[scopes.worker]
secrets = ["DATABASE_URL", "QUEUE_TOKEN"]
```

```bash
secretspec run   --scope api    -- ./api
secretspec check --scope worker
secretspec export --scope api --format dotenv
```

`--scope` (env `SECRETSPEC_SCOPE`) is on `run`, `check`, and `export`.

## The model

`visible = effective_profile ∩ scope`, `accessed = transitive_dependency_closure(visible)`. Resolution plans `accessed` so an in-scope composed secret can build its value; every output surface is then narrowed back to `visible`.

- Membership-only: a scope never changes `required`/`default`/providers/references/generation/`as_path` or the `{project}/{profile}/{key}` storage address. No per-scope overrides, so there is no profile × scope precedence.
- No scope → the complete profile, unchanged. Filtering happens at `resolve_profile_secret_names`, the single worklist upstream of validation, planning, fetching, prompting, generation, and audit, so all of them scope at once and an unknown scope fails **before any provider is touched**.
- A scope naming a secret the profile does not declare yields a smaller set rather than an error, so a scope can be reused across profiles. A required secret the scope excludes does not block resolution.
- **Composed secrets resolve their inputs without exposing them.** An in-scope `DATABASE_URL` fetches `DB_USER`/`DB_PASSWORD` and drops them from the value map, `as_path` files, report, and missing/default lists. A secret neither visible nor a dependency of one is never fetched; an empty scope contacts no provider.
- **`run --scope`** scrubs every manifest-declared secret outside the visible set — across *all* profiles, via `Command::env_remove` — so a value inherited from a parent shell under another profile cannot leak into the child. As you said, secret minimization, not an authorization boundary; noted in the docs.
- **Typed loaders** ignore an ambient `SECRETSPEC_SCOPE` and always resolve the full profile. Option 1 as you directed: the macro and generated API are unchanged.
- **Active scope** is surfaced in the resolve response, the report, and `check --explain`.
- Under project `extends`, a child scope **replaces** the parent's of the same name outright — no implicit union, since replacement is the safe default for an allowlist. Documented, with a test.

## Scopes × presence constraints (new — #184 landed after review)

Presence constraints arrived after the last review and need a defined interaction. A group is judged over the members the scope **exposes**:

- No visible member → not that consumer's concern, not enforced.
- Some visible members → enforced over those alone, so a scope never inherits a guarantee resting on a secret it hides. `at_least_one = "cloud"` satisfied profile-wide by `GCP_KEY` still **fails** a scope showing only an absent `AWS_KEY`.
- `exactly_one` stays enforced whenever two visible members are both present — scoping narrows what is judged, never whether it is judged.
- A secret fetched only as a hidden composition input never counts as present, and a violation names only visible members.

The rejected alternative — skipping constraints under a scope, which the existing partial-plan guard would have done silently — would turn a safety check off exactly when a service is being narrowed. `get`'s own partial-plan guard is untouched.

## The three previously-open questions, answered

1. **Audit records `accessed`.** The log captures provider access, and a hidden composition input *is* read. `run` still records the visible set, because it logs what was injected — two events, two facts.
2. **`as_path` dependency of a visible composition was a bug**, not a preference: an `as_path` value *is* its temp-file path, so filtering temp files by the visible set handed the consumer a dangling path. Temp files are no longer filtered — every accessed non-visible secret is by construction a dependency of a visible one. Consistent with composition already carrying its inputs' content in derived form; the input stays out of the environment.
3. **Visible-but-optional-missing secrets stay unscrubbed.** They are *inside* the visible set, so a parent-exported value is inherited exactly as it would be unscoped. Scoping decides which secrets are in play, not the semantics of one it admits.

## Tests & docs

590 green; `fmt`/`clippy` clean. Coverage includes composed output without dependency exposure, nested composition, cross-profile inherited-env scrubbing, empty-scope/no-provider behavior, typed-loader behavior with `SECRETSPEC_SCOPE` set, scoped prompting never naming a hidden dependency, the four constraint-interaction cases, the `as_path` fix, and the audit decision.

Two bugs found and fixed along the way: interactive `check --scope` prompted for (and overwrote) a composed secret's out-of-scope dependencies; and a doc claimed typed loads can be scoped via the builder, which they cannot.

Rebased on current `main`. Composition syntax moved from `{NAME}` to `${NAME}` in 09df71a after this branch was cut — tests updated — and the CHANGELOG entry moved out of the now-released 0.16.0 section, with all version markers retargeted to **0.17**.

Deferred as you asked: multiple scopes, scope-specific overrides, per-secret tags, extendable profiles, scope-aware generated types.


ojhermann:
@domenkozar — draft, but there's one decision I'd like your steer on before I finalize, and it's the only thing blocking this from being review-ready.

**The config + CLI surface is done and follows your #137 design exactly** (membership-only, intersection semantics, no per-scope overrides, and the `run --scope` env scrub that removes excluded parent-exported secrets from the child). Tests are green.

**The open question is the derive macro.** `declare_secrets!` reads `secretspec.toml` at compile time, but `--scope` is a runtime selection, so the generated type can't just be "the scope's fields." I took **option 1**: scope is a resolution-time filter only, macro unchanged, typed SDK unaffected — which fully covers the CI/devenv motivating case. The alternatives (macro emits a union with scope membership, or one struct per scope) are additive and I'll build whichever you prefer before this leaves draft.

Which way would you like it pointed? That's the one answer I need to take this out of draft.

domenkozar:
I think we have to be careful here, there are many things at play and I'd like to spend a tiny bit more time exploring all the options we have before we commit to a solution.

Profiles are kind of mixing environments and scopes, while having only one level of inheritance.

It would be good to come up with ideal design and then see if we can fit it into the current one.

Sorry for dragging you into this one and then pulling the plug, but I'd rather do it once than twice.

domenkozar:
Thanks, this is pointed in the right direction. On the macro question: let’s take option 1 for now—keep the generated typed API unchanged and treat scopes as a runtime resolution/output feature. I don’t want per-scope structs or profile × scope generated types in this PR.

Before taking it out of draft, please rebase onto current `main` and account for composed secrets. We need to distinguish the consumer-visible set from the internal resolution set:

```text
visible = effective_profile ∩ scope
accessed = transitive_dependency_closure(visible)
```

Resolve `accessed`, but inject/export only `visible`. For example, a scope containing a composed `DATABASE_URL` should fetch `DB_USER` and `DB_PASSWORD` without exposing those inputs to the child.

A few related correctness points:

- `run --scope` should scrub every manifest-declared secret outside the visible set, not only excluded names from the selected profile. Otherwise a secret inherited from a parent environment under another profile can leak through.
- “Macro unchanged” should also mean typed loaders are not accidentally affected by ambient `SECRETSPEC_SCOPE`; currently they call `Secrets::load()` and expect the full generated shape.
- A valid empty scope/intersection should not initialize or contact a provider.
- Please make the active scope explicit in untyped resolver/report output if those APIs support scoped resolution.
- Scope replacement under project `extends` should remain whole-value replacement rather than implicit union; please document that, since replacement is safer for an allowlist.

Tests I’d especially like to see are: composed output without dependency exposure, nested composition, cross-profile inherited-env scrubbing, empty-scope/no-provider behavior, and typed-loader behavior with `SECRETSPEC_SCOPE` set.

Let’s keep multiple scopes, scope-specific overrides, per-secret tags, extendable profiles, and scope-aware generated types out of this PR. One membership-only scope per resolution is enough for the first version.

ojhermann:
Agree with pausing here — better to get the model right once than ship a `--scope` we have to walk back. And I think your diagnosis is the right frame: a profile today does two independent jobs — the **environment** axis (how each secret resolves: `required`/`default`/provider chain) and the **membership** axis (which secrets are in play) — and with one level of inheritance that forces a cross-product (`prod`, `prod-api`, `prod-worker`, …). Scopes as they stand are really just the *extraction* of that membership axis (`resolved = intersection(profile, scope)`, never touching a secret's resolution rules), so I read this PR less as "a third concept" and more as "letting profiles go back to meaning only environment." But you're right that it leaves real questions open — laying them out:

**1. How membership is represented (the real fork).**
- *This PR:* a separate `[scopes]` table listing member secrets — centralized and easy to audit ("what does `api` need?" reads in one block), but it can drift from the profile decls, which is why it needs the "scope references a secret not declared in any profile" validation.
- *Alternative:* tag membership on each secret (`API_KEY = { …, scopes = ["api", "worker"] }`) — drift-proof by construction and composes naturally, but "what does `api` need" becomes a cross-secret grep, and untagged secrets need a default policy.

These two are duals of the same bipartite graph, so we can pick the ergonomics without changing the resolution model.

**2. Composition / inheritance depth.** Neither profiles (one level) nor scopes (none) compose today. I'd keep v1 flat, but pick syntax that leaves room for scope-includes-scope and/or multi-level profile inheritance so we don't have to break it later.

**3. Intersection semantics.** When a scope names a secret the active profile doesn't declare, the current behavior silently drops it — so `--scope api` can quietly resolve to fewer secrets than expected. Silent-drop vs. hard-error vs. warn is a real UX call worth making deliberately.

My lean: keep the axis this PR establishes (rather than deepening profile inheritance, which just makes the cross-product cheaper while keeping the overload), decide #1 explicitly, and consciously defer #2/#3 — but happy to go whichever direction you think fits best.

ojhermann:
Just saw your detailed reply landed while mine was in flight — crossed in the mail, and we're aligned. Taking option 1: single membership-only scope, `[scopes]` table, macro/generated API untouched, tags/multi-scope/extendable-profiles deferred. I'll rebase onto `main` and build the `visible` vs `accessed` (dependency-closure) split so composed secrets resolve their inputs without exposing them, scrub everything outside `visible` in `run --scope`, keep typed loaders free of `SECRETSPEC_SCOPE`, short-circuit empty scopes before any provider contact, and surface the active scope in untyped output — with the tests you listed.

ojhermann:
@domenkozar — rebased onto `main` (composed secrets included) and implemented the spec. Summary of what changed; still draft pending your answers on the three questions below.

**What changed**
- **Scoped composition:** an in-scope composed secret (e.g. `DATABASE_URL`) resolves its out-of-scope inputs (`DB_USER`/`DB_PASSWORD`) to build its value, but they're filtered from every output — value map, `as_path` temp files, report, missing/default lists. A secret that's neither visible nor a dependency of one is never fetched.
- **`run --scope`** scrubs every manifest-declared secret outside the visible set across *all* profiles, even ones the parent exported — so nothing inherited from another profile leaks into the child.
- **Empty scope** contacts no provider.
- **Typed loaders** ignore an ambient `SECRETSPEC_SCOPE` and always resolve the full profile; the untyped CLI/SDK path still honors it.
- **Active scope** is surfaced in the resolve response and report (optional field, omitted when unscoped) and in `check --explain`.
- **`extends`:** a child scope replaces the parent's of the same name outright (no union) — documented, with a test.
- Tests for all of the above.

**Two bugs found and fixed along the way**
- Interactive `check --scope` prompted for a composed secret's out-of-scope dependencies — printing their names and overwriting already-present values. Prompting is now limited to the visible set.
- A doc claimed typed loads can be scoped via the builder; they can't (no such method). Corrected.

**Three questions for you** — didn't want to decide these unilaterally:
1. **Audit:** a scoped `run` reads out-of-scope dependencies to build a composition, but the audit log records only the *visible* names. Record `visible` (fewer names disclosed) or `accessed` (faithful to what was read)? I lean `accessed`.
2. **`as_path` dependency of a visible composition:** the composed value embeds the dependency's temp-file path, but under a scope that file is dropped, leaving a dangling path. Reject the config, keep the file, or leave as-is?
3. **Visible-but-optional-missing secrets** aren't scrubbed from `run` (the scope admits them; they just didn't resolve) — deliberate, but flag if you'd rather scrub them.

ojhermann:
@domenkozar — rebased onto current `main` and decided the three questions I'd left open, so this is a concrete proposal rather than a queue of asks. Reasoning below in case you'd steer any of them differently; all three are cheap to flip.

**A fourth question appeared during the rebase, and it's the one worth your attention.** Presence constraints (#184) landed after your review, and scopes have to define an interaction with them. What I implemented:

- A group is judged over the members the scope **exposes**. A group with no visible member is not that consumer's concern and is not enforced.
- A group with visible members is enforced over those alone, so a scope never inherits a guarantee that rests on a secret it hides. If `at_least_one = "cloud"` is satisfied profile-wide by `GCP_KEY`, a scope showing only an absent `AWS_KEY` **fails** — the alternative would be a scoped service passing `check` on the strength of a credential it cannot see.
- `exactly_one` stays enforced whenever two visible members are both present. Scoping narrows *what* is judged, never *whether* it is judged — the mutual-exclusion safety property is the whole point of that mode.
- A secret fetched only as a hidden composition input never counts as present, and a violation message names only visible members, preserving the invariant that a scope never discloses a secret it hides.

The alternative I rejected: skipping constraints under a scope entirely. There's a precedent for it — `get`'s deliberately partial plan already skips them — and the existing length guard would have done it silently, which is how I noticed. But a scoped resolution isn't a least-access read of one secret; it's whole-profile validation of a declared subset, and silently dropping `exactly_one` there would turn a safety check off exactly when a service is being narrowed. So scoped resolution is opted back in explicitly, and `get`'s guard is untouched.

## The three questions, answered

**1. Audit: `visible` or `accessed`?** Went with **`accessed`**, as I leaned. The log's purpose is to record what was read from a provider, and a hidden composition input *is* read — recording only visible names understates provider access, which is the one thing the log exists to capture. The disclosure argument doesn't hold up: the audit log is a local artifact written by the process that already holds the credentials, not something handed to the scoped consumer. `run` still records the **visible** set, because it logs what was injected. Two events, two different facts.

**2. `as_path` dependency of a visible composition.** This turned out to be a **bug**, not a preference — thank you for pointing at it. An `as_path` secret's resolved value *is* its temp-file path, so a visible composition built from a hidden `as_path` input embeds that path in its own value. Filtering temp files by the visible set deleted the file and handed the consumer a path to nothing. Fixed by not filtering temp files: every accessed non-visible secret is by construction a dependency of a visible one, so its file is exactly what the composition needs. That's consistent with composition's existing contract — a composed DSN already carries its inputs' content in derived form — while the input itself stays out of the environment. There's a test that runs a child, reads the embedded path, and asserts the certificate is still there.

**3. Visible-but-optional-missing secrets aren't scrubbed from `run`.** Kept as-is, and I think the framing makes it uncontroversial: such a secret is *inside* the visible set, so a parent-exported value is inherited exactly as it would be with no scope active. Scoping decides which secrets are in play, not the semantics of one it admits. Scrubbing it would make `--scope` quietly stricter than the unscoped path for a secret the scope explicitly names.

## Rebase notes

- Composition syntax moved from `{NAME}` to `${NAME}` in 09df71a after this branch was cut; the scope tests are updated.
- The scopes CHANGELOG entry had landed in what is now the released 0.16.0 section — moved to Unreleased, and all "0.16" version markers in the configuration, CLI, and inheritance docs retargeted to **0.17**.

## Tests & docs

590 green; `fmt`/`clippy` clean. New coverage for the constraint interaction (visible-member enforcement, unenforced when no member is visible, `exactly_one` with both visible, hidden input not counting as present), the `as_path` fix, and the audit decision. All of the above is documented in the configuration reference — the scope behavior list and a reciprocal note in the presence-constraints section.

Still your call on the four decisions; happy to flip any of them.

domenkozar:
Hey, why did you close this?

ojhermann:
I think it happened automatically when I closed my fork.

domenkozar:
Do yoj want to reopen the PRs?

ojhermann:
I can get them back open a bit later today.

PR #24, opened 2025-07-23
Add Bitwarden provider with comprehensive security enhancements

ashebanow:
This PR adds a new Bitwarden provider supporting both Bitwarden Password Manager and Bitwarden Secrets Manager.

Closes #15

## Features
- **Bitwarden Password Manager support** via `bitwarden://` URIs
- **Bitwarden Secrets Manager support** via `bws://` URIs  
- **Comprehensive error message sanitization** to prevent sensitive data exposure
- **Cross-platform timeout handling** with proper resource cleanup
- **Extensive test coverage** with 18 passing tests

## Security Enhancements
- Enhanced pattern detection for JWT tokens, Windows file paths, and extended secret field names
- Fixed timeout resource leaks with proper thread and process cleanup
- Sanitization applied to all error paths preventing sensitive information exposure

## Implementation Details
- Thread-safe timeout implementation using Arc<Mutex<Option<Child>>> pattern
- Support for multiple item types (login, card, identity, SSH key, secure note)
- Configurable CLI timeouts and paths
- Performance logging with SECRETSPEC_PERF_LOG environment variable

All tests pass and the implementation follows the existing provider patterns and architecture.

Issue #15, opened 2025-07-21
Req to Support Bitwarden


jkfnc:

vinnymeller:
I took a shot at this but realized I can't really finish or test it all on my own, but maybe can be useful as a starting point or reference for someone else: https://github.com/cachix/secretspec/commit/3def4ed9c7949a5b43f25ce597c786abc171c628


It *does*:
- work with a bitwarden password manager personal account

It *doesn't*:
- Take into account folders at all, and I assume is broken for orgs/collections

Untested:
- self-hosted server
- There is some code for collections is probably broken as I couldn't test any of it

Also, I didn't even know that bitwarden had a proper [secret manager](https://bitwarden.com/go/secrets-manager/). My provider is for the password manager. Not sure if those would be separate providers, if they could fit into one, or what 🤷

ashebanow:
I'm going to take a crack at this as well.

Testing with Vaultwarden isn't really practical for me, but I do use both Bitwarden and the Bitwarden Secrets Manager, so I can test with those.

domenkozar:
Support for BWS is in main!

bsorescu:
Since BWS support landed (thanks!), the personal-vault half of this issue still seems open — I'd like to take a crack at it, and I'd rather align on the approach before writing code, given how #24 went.

Two viable designs:

1. **`bw serve` REST API** (my preference): the provider talks to the CLI's local Vault Management API over HTTP (async reqwest, same shape as `vault.rs`). No subprocess spawn per secret, no `BW_SESSION` plumbing — lock/unlock state lives in the `bw serve` process. Downside: the user needs `bw serve` running (the provider can spawn/manage it lazily).
2. **CLI subprocess** (`bw get item` per secret, same pattern as `pass.rs`/`lastpass.rs`): simpler, but slow (one `bw` process per secret; the CLI has known multi-second startup) and needs session-key handling.

Scope: personal vault items as secure notes with custom fields (one item per project/profile, matching how Kamal's bitwarden adapter addresses them), `get`/`set`, self-hosted servers supported via the CLI's own `bw config server`. Small, focused, tests + the full docs checklist from CLAUDE.md.

Would you take a PR along the lines of (1)? Happy to adjust the design to whatever you'd prefer to maintain.

(Context: I run this stack daily — secretspec with keyring + a bw→keyring sync script today — and just filed #140 while building it.)

ashebanow:
[...]

PR #166, opened 2026-07-17
feat: add Bitwarden Password Manager provider

ashebanow:
Adds a Bitwarden **Password Manager** provider (`bw://`) backed by the official `bw` CLI.

This resumes the effort from #24 — the renewed discussion in #15 is what inspired picking this work back up. It starts from the Password Manager core of that PR and reworks it for today's codebase.

## What's included

- `bw://` provider shelling out to the `bw` CLI, with vault-wide access to all item types: logins, cards, identities, SSH keys, and secure notes
- Provider id `bw` matches the CLI it drives and the cargo feature, mirroring how `bws://` is named after the `bws` CLI
- Sensible per-type default fields, plus explicit selection via URI query params (`bw://?type=login&field=username`) and `BITWARDEN_DEFAULT_TYPE` / `BITWARDEN_DEFAULT_FIELD` env vars (prefix kept clear of the `bw` CLI's own `BW_*` namespace)
- `ref = { item = "...", field = "..." }` mapping in `secretspec.toml`, so secret keys stay valid identifiers while pointing at arbitrarily named vault items
- Collection/organization addressing (`bw://org@collection`) and self-hosted servers (`?server=`)
- Item creation and update on `secretspec set`
- Curated error mapping for common `bw` failure modes: CLI not installed (with install instructions), not logged in, vault locked; secret values pass via stdin and the session via the environment, never argv
- `bw` cargo feature (default-enabled; adds only a `base64` dependency since the provider drives the CLI rather than linking an SDK)
- Docs: provider page (`docs/providers/bw.md` with 0.16 version notice) and all provider listings per the adding-providers checklist — sidebar, llms.txt description, concepts/reference tables, security table, landing page, quick-start, README — each labeled `(0.16+)`
- CHANGELOG entry under Unreleased, unit tests for URI/config parsing
- `tests/bitwarden_integration.sh`: end-to-end suite against a real vault covering all five item types, `ref` mapping, env-var defaults, error paths, and set/update; auto-creates and cleans up its own test items
- `bitwarden-cli` added to devenv for integration testing

## Changes relative to #24

- **Secrets Manager support dropped** — upstream now ships a dedicated `bws://` provider, so this PR is Password Manager only
- Ported to the current `ProviderUrl`/`Address` API and provider registration
- #24's custom timeout machinery, error-message sanitizer, and perf instrumentation were **not** carried over — the rewrite follows the same straightforward CLI-execution pattern as the 1Password/LastPass providers
- Integration script hardened: automatic test-item setup/teardown and valid-identifier keys via `ref` mapping
- The integration script's header documents a vaultwarden-based path to fully automated, credential-free testing (future work)

## Testing

- `cargo test --all` green locally (macOS)
- Integration script run against a real vault: all item types, `ref` mapping, env-var defaults, error handling, and set/update paths

## Notes for reviewers

- Provider listings are complete per the adding-providers checklist (sidebar, concepts/reference tables, landing page, quick-start, README), all labeled 0.16+
- Real-world testing so far has focused on a personal vault exercising all five item types; organization/collection layouts have had lighter coverage — extra scrutiny welcome there
- The integration script's header documents a vaultwarden-based path to fully automated, credential-free testing, left as future work

Refs #15 (inspiration for resuming), supersedes #24.

AI Note: this project's code and documentation were produced with the help of UI. The old pull request was developed with the help of Claude Code and Claude Opus/Sonnet, the additional work for this pull request was done using my custom pi environment and Deepseek v4 Pro/Flash.


bsorescu:
Vaultwarden testing done, as promised on #15: the full `tests/bitwarden_integration.sh` suite passes **19/19** against a disposable Vaultwarden (`vaultwarden/server:latest`) through the genuine `bw` CLI 2026.6.0 — reads across all item types (login, card, identity, SSH key, secure note, custom fields), writes (`set` creating login/card items), and the error-handling cases.

And since you asked about automating it: I've opened ashebanow/secretspec#2 against your branch implementing the FUTURE WORK plan from the script header — disposable Vaultwarden + fixture account, no repository secrets, fork-PR-friendly. Two gotchas discovered on the way, handled in the harness:

- the 2026+ `bw` CLI refuses plain `http://` servers outright, so the harness fronts Vaultwarden with a Caddy TLS proxy (internal self-signed cert + `NODE_TLS_REJECT_UNAUTHORIZED=0` scoped to the fixture run);
- `bw` has no `register`, so a small `vaultwarden_bootstrap.py` implements the client-side registration crypto (PBKDF2 → HKDF stretch → EncString type 2 → RSA keypair) against the identity API.

`BITWARDENCLI_APPDATA_DIR` keeps the whole thing isolated from the developer's real bw config. A GitHub Actions job wiring this up should be a ~20-line follow-up once this lands.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+                config.organization_id = Some(url.username());
+                config.collection_id = Some(host);
+            } else {
+                // Just collection ID
+                config.collection_id = Some(host);
+            }
+        }
+
+        // Parse query parameters
+        for (key, value) in url.query_pairs() {
```
domenkozar, line 432: **P1 — Enable query parsing in bw-only builds**

When consumers build only the advertised feature with `cargo check -p secretspec --no-default-features --features bw`, this call fails because `ProviderUrl::query_pairs` is not compiled unless another provider feature or `test` is enabled. Please include `bw` in that method’s cfg gate so the feature compiles independently.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+    /// Returns specific errors for:
+    /// - Missing Bitwarden CLI installation
+    /// - Authentication required (not logged in or unlocked)
+    /// - Command execution failures
+    fn execute_bw_command(&self, args: &[&str]) -> Result<String> {
+        let mut cmd = Command::new("bw");
+
+        // Configure server if specified
+        if let Some(server) = &self.config.server {
+            cmd.env("BW_SERVER", server);
```
domenkozar: **P1 — Configure self-hosted servers through a supported bw mechanism**

For `bw://?server=...`, setting `BW_SERVER` has no effect because the official `bw` CLI does not read that environment variable; `BW_SERVER=https://example.invalid bw status` still reports the configured default server. Consequently every advertised self-hosted operation targets the CLI’s existing server instead, so this needs a supported configuration mechanism.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+        if !self.is_authenticated()? {
+            return Err(SecretSpecError::ProviderOperationFailed(
+                "Bitwarden authentication required. Please run 'bw login' and 'bw unlock', then set the BW_SESSION environment variable.".to_string(),
+            ));
+        }
+
+        // Use Bitwarden's built-in search to find items matching the key
+        let mut list_args = vec!["list", "items", "--search", item_name];
+
+        // Add organization filter if configured (from config or environment variable)
```
domenkozar: **P1 — Apply the configured collection to vault searches**

When `bw://collection-id`, `bw://org@collection`, or `BITWARDEN_COLLECTION` is used, this lookup applies only the organization filter and ignores the collection. If another collection contains the same item name, reads can return its value, and the equivalent unfiltered write search can update it. Please restrict searches with the configured collection ID.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+}
+
+impl BitwardenFieldType {
+    /// Convert from integer to enum
+    pub fn from_u8(value: u8) -> Option<Self> {
+        match value {
+            0 => Some(BitwardenFieldType::Text),
+            1 => Some(BitwardenFieldType::Hidden),
+            2 => Some(BitwardenFieldType::Boolean),
+            _ => None,
```
domenkozar, line 115: **P1 — Accept linked Bitwarden custom fields**

Bitwarden defines custom-field type `3` for linked fields, but this parser rejects it. Since `set_to_password_manager` deserializes every item returned by `bw list items`, one linked field anywhere in the vault causes every write to abort with `Unknown field type: 3`; reads of an item containing one also fail.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+            "username": null,
+            "password": null,
+            "totp": null,
+            "uris": []
+        });
+
+        match field.to_lowercase().as_str() {
+            "username" => login_data["username"] = serde_json::Value::String(value.to_string()),
+            "totp" => login_data["totp"] = serde_json::Value::String(value.to_string()),
+            _ => login_data["password"] = serde_json::Value::String(value.to_string()),
```
domenkozar: **P1 — Preserve explicitly named custom fields when creating items**

When a new login is created with a documented custom field such as `field=api_key`, this fallback stores the value in `login.password`. A subsequent read using the same provider looks for the custom `api_key` field and returns nothing. Card and Identity creation use the same incorrect fallback pattern for unknown field names.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+        item: &BitwardenItem,
+        target_field: Option<&str>,
+        value: &str,
+    ) -> Result<()> {
+        // Determine which field to update: explicit > env > config > smart default
+        let field = target_field
+            .map(|s| s.to_string())
+            .or_else(|| std::env::var("BITWARDEN_DEFAULT_FIELD").ok())
+            .or_else(|| self.config.default_field.clone())
+            .unwrap_or_else(|| "password".to_string());
```
domenkozar: **P1 — Choose type-specific defaults when updating existing items**

For an existing Secure Note, Card, Identity, or SSH-key item without an explicit field, this always selects `password`. The updater therefore creates a custom password field while the corresponding getter defaults to notes/value, card number, email, or private key, so `set` succeeds but a following `get` returns the old value.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,1757 @@
...
+            item_json["fields"] = serde_json::Value::Array(vec![]);
+        }
+
+        let fields = item_json["fields"].as_array_mut().ok_or_else(|| {
+            SecretSpecError::ProviderOperationFailed("Invalid fields array".to_string())
+        })?;
+
+        // Look for existing field
+        for field_obj in fields.iter_mut() {
+            if field_obj["name"].as_str() == Some(field) {
```
domenkozar: **P2 — Match custom-field names consistently when updating**

Reads match custom-field names case-insensitively, but updates use case-sensitive equality here. With an existing field named `API_KEY` and a reference to `api_key`, reading works while setting adds a duplicate field; later reads encounter the original field first and continue returning the stale value.

docs/src/content/docs/providers/bw.md
```diff
@@ -0,0 +1,200 @@
+---
+title: Bitwarden Password Manager Provider
+description: Bitwarden Password Manager secrets management integration
+---
+
+The `bw` provider integrates with Bitwarden Password Manager for secret management with vault-wide access to all item types.
+
+:::note[Version compatibility]
```
domenkozar, line 8: **P2 — Advertise the actual target release everywhere**

The base already contains the released `0.16.0` section dated 2026-07-17, so this unreleased provider cannot be described as an upcoming 0.16 feature. Please update every new listing to the actual target release, including the homepage entries that currently omit a version entirely; otherwise users of 0.16 are told the unavailable provider is supported.

docs/src/content/docs/providers/bw.md
```diff
@@ -0,0 +1,200 @@
...
+```bash
+# Get value from secure note
+$ secretspec get 'Legacy Config' --provider 'bw://?type=securenote&field=config_value'
+```
+
+### Profile Configuration
+
+```toml
+# secretspec.toml
+[development]
```
domenkozar: **P2 — Use valid profile configuration in the provider guide**

These tables do not match the project configuration schema: profiles are under `[profiles.<name>]`, and profile provider selection is configured through the profile’s `defaults.providers` alias list rather than a `provider` string. Users following this example will get an invalid or misinterpreted `secretspec.toml` instead of collection-specific profiles.

CHANGELOG.md
```diff
@@ -51,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.16.0] - 2026-07-17
 
 ### Added
+>>>>>>> upstream/main
```
domenkozar: **P2 — Remove the unresolved changelog merge marker**

The literal `>>>>>>> upstream/main` is a merge artifact and will be published as part of the 0.16 changelog unless removed.

bsorescu:
Following up on the review: the four findings that are reproducible against a live vault are now regression tests in the Vaultwarden harness (added to ashebanow/secretspec#2) — one test per finding, each reporting `REPRODUCED`/`FIXED`, exiting green only when all are fixed.

Current run against disposable Vaultwarden confirms every one of them verbatim:

```
── R1: linked custom field (type 3) poisons unrelated writes ──
  REPRODUCED R1 — any write aborts while a linked field exists (Unknown field type: 3)
── R2: named custom field preserved when creating a new item ──
  REPRODUCED R2 — value not readable via field=api_key after create (stored in login.password instead)
── R3: type-specific default on update (secure note) ──
  REPRODUCED R3 — set wrote a password field; get still returns 'old-note-value'
── R4: case-insensitive field matching on update ──
  REPRODUCED R4 — update added a duplicate field; get still returns 'old-value'

Findings fixed: 0 · reproduced: 4
```

So @ashebanow has an executable checklist for those four (`RUN_REGRESSIONS=1 tests/vaultwarden_harness.sh`) — they'll flip to FIXED as the fixes land. The collection-filtering finding needs an org + two collections in the fixture account; happy to extend the bootstrap for that as a follow-up if useful. On the self-hosted config finding: the only supported mechanism the CLI offers is `bw config server` (which requires being logged out), so the provider likely needs to shell that out — or document it as a user prerequisite, which is what Kamal's bitwarden adapter effectively does.

ashebanow:
I'm traveling this long weekend, not sure how much time I'll have. But b as
CK Tuesday at the latest.

domenkozar:
@ashebanow  I'm planning to do a 0.17 release on Monday if you'd like to get this in. No issue if we target 0.18.

nmunnich:
Fwiw this is the provider I am missing to "complete the puzzle" for me, so I am very very much looking forward to this landing. Cheers to all involved.

ashebanow:
I'm at a family event this weekend, let's target 0.18, thanks!

ashebanow:
OK, a large set of changes incoming, just waiting for the integration tests to pass after merging latest from this repo. Before it arrives, I want to thank @bsorescu for the vaultwarden help, and @domenkozar for his fantastic code review. Its clear to me that I was too trusting of my AI code to do the merge of all of secretspec's changes since the last time the code was submitted, and that I had not paid enough personal attention to the changes in rules for providers. The platform has really matured impressively.

My response to all those review comments: add tests, and more tests. Those tests in turn found more issues, which have been addressed. I had Claude Opus 5 do a detailed code review as well, looking at the code with a strong critical eye. Even so, there are holes: the biggest being that the vaultwarden integration tests are still required to be run manually, primarily since the setup requirements are still not sufficiently automated to allow for regular, unattended runs IMHO. We should file a separate bug to improve and integrate these tests into the larger integration test process. There is also a vaultwarden issue with ssh entries that we found and will report to vaultwarden.

docs/src/content/docs/providers/bw.md
ashebanow: moved to 0.18

CHANGELOG.md
ashebanow: removed

secretspec/src/provider/bw.rs
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

docs/src/content/docs/providers/bw.md
ashebanow: fixed

secretspec/src/provider/bw.rs
ashebanow: fixed

bsorescu:
Independent verification of the review fixes, on `9ea1a179` against a disposable Vaultwarden (macOS, bw CLI 2026.6.0, `vaultwarden/server:latest`):

- **Regression tests (`RUN_REGRESSIONS=1`): 4/4 FIXED** — R1 linked-field poisoning, R2 field lost on create, R3 update/read default mismatch on non-login items, R4 case-sensitive update matching. All previously REPRODUCED on `857e32d4`, all green now.
- **Integration suite: 39/39 passed** (including the new create → update → read-back sweep).
- **Collection addressing: 11/11 passed** — the org + collections fixture bootstraps cleanly.

Full harness exit 0. Nice work — from where I sit this is ready for 0.18.

Two environment findings from the runs, in case they're worth a small follow-up:

1. **The suite doesn't isolate the developer's global secretspec config.** With a `[defaults] profile = "..."` in `~/.config/secretspec/config.toml` (XDG path on macOS too), every `set`/`get` fails with `Invalid profile: '...' is not defined in secretspec.toml`, which cascades into all findings reporting REPRODUCED and the integration suite failing wholesale. Suggested fix, symmetric with what the harness already does for `BITWARDENCLI_APPDATA_DIR`: export `XDG_CONFIG_HOME` to a tmpdir in `vaultwarden_harness.sh`. Happy to send that as a small PR to your branch if useful.
2. **The org fixture needs a recent Vaultwarden image.** My local `vaultwarden/server:latest` was ~2 months stale and its `/api/accounts/profile` response has no public key under any of the probed locations (only `key`/`privateKey`), so step 4 aborts with `no publicKey in profile response`. A `docker pull` fixed it — might deserve a one-line note in the harness header.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+        // widen the search rather than narrow it.
+        let filter = self.search_filter_args()?;
+        list_args.extend(filter.iter().map(String::as_str));
+
+        let output = self.execute_bw_command(&list_args)?;
+        let items: Vec<BitwardenItem> = serde_json::from_str(&output)?;
+
+        // If we found items, use the first one
+        if let Some(item) = items.first() {
+            return self.extract_value_from_item(item, field_hint);
```
domenkozar, line 1497: **P1 — Select an exact item of the requested type before reading** — `bw list items --search` performs broad string matching and can return multiple item types, but this always extracts from the first result. A request for `API_KEY` can therefore read `API_KEY_OLD`, and `bw://?type=card` cannot disambiguate a Card from a same-named Login because the configured type is never consulted. Filter by exact name and any explicitly selected type before extracting.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+        if let Some(item) = items.iter().find(|item| item.name == item_name) {
+            return self.update_existing_item(item, target_field, value.expose_secret());
+        }
+
+        // Strategy 2: Contains item_name in name (case-insensitive)
+        if let Some(item) = items
+            .iter()
+            .find(|item| item.name.to_lowercase().contains(&item_name.to_lowercase()))
+        {
+            return self.update_existing_item(item, target_field, value.expose_secret());
```
domenkozar, line 1877: **P1 — Restrict updates to an exact item match** — When no exact item exists, setting `API_KEY` while `OLD_API_KEY` is present takes this substring branch and overwrites the old item instead of creating `API_KEY`. Since item names are provider coordinates, fuzzy matching on a write can silently corrupt an unrelated secret and should not select an update target.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+        // If specific field requested, check custom fields first
+        if let Some(field_name) = resolved_field
+            && let Some(value) = self.extract_from_custom_fields(item, field_name)?
+        {
+            return Ok(Some(SecretString::new(value.into())));
+        }
+
+        // Look for legacy "value" field (backward compatibility)
+        if let Some(value) = self.extract_from_custom_fields(item, "value")? {
+            return Ok(Some(SecretString::new(value.into())));
```
domenkozar, line 1621: **P1 — Stop falling back after an explicit secure-note field misses** — When an explicit Secure Note field is absent, this falls through to the legacy `value` field and then the note body. For example, `field=config_value` can return an unrelated `value`, and `field=notes` returns `value` whenever that custom field exists. An explicit selector should return only that field or `None`, rather than another secret.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+        match field.to_lowercase().as_str() {
+            "code" | "cvv" | "cvc" => {
+                card_data["code"] = serde_json::Value::String(value.to_string())
+            }
+            "cardholder" | "name" => {
+                card_data["cardholderName"] = serde_json::Value::String(value.to_string())
+            }
+            "brand" => card_data["brand"] = serde_json::Value::String(value.to_string()),
+            "number" => card_data["number"] = serde_json::Value::String(value.to_string()),
+            _ => {
```
domenkozar: **P1 — Preserve supported built-in fields when creating typed items** — For a new Card, `field=exp_month` or `field=exp_year` falls into the custom-field branch, while the getter recognizes those names as built-ins and reads the still-null `card.expMonth` or `card.expYear`; `set` succeeds but an immediate `get` returns nothing. `identity_template` has the same mismatch for `first_name` and `last_name`. Creation should recognize the same built-in aliases as reading and updating.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+            uri.push_str(&ProviderUrl::encode(org_id));
+            uri.push('@');
+        }
+        if let Some(ref coll_id) = self.config.collection_id {
+            uri.push_str(&ProviderUrl::encode(coll_id));
+        }
+        if let Some(ref server) = self.config.server {
+            uri.push('?');
+            uri.push_str(&format!("server={}", server));
+        }
```
domenkozar, line 2559: **P2 — Include behavior-changing query options in the canonical URI** — A provider constructed from `bw://?type=card&field=api_key` reports only `bw://` because `uri()` emits `server` but drops `type` and `field`. SecretSpec uses this URI for provenance and cached-route fingerprints, so changing the selected field can leave an old cached value considered valid and reports cannot distinguish the providers.

secretspec/src/provider/bw.rs
```diff
@@ -0,0 +1,3209 @@
...
+            match key.as_ref() {
+                "org" | "organization" => config.organization_id = Some(value.into_owned()),
+                "collection" => config.collection_id = Some(value.into_owned()),
+                "server" => config.server = Some(value.into_owned()),
+                "folder" => config.folder_prefix = Some(value.into_owned()),
+                "type" => {
+                    if let Some(item_type) = BitwardenItemType::from_str(&value) {
+                        config.default_item_type = Some(item_type);
+                    }
+                }
```
domenkozar: **P2 — Reject unsupported item types instead of defaulting to Login** — An invalid value such as `bw://?type=sshkee` is silently ignored, leaving the default Login type. A subsequent `set` therefore creates a Login item despite the requested type rather than reporting the configuration typo.

tests/bitwarden_integration.sh
```diff
@@ -0,0 +1,519 @@
...
+
+    # Check if item already exists
+    local existing_id
+    existing_id=$(BW_SESSION="$BW_SESSION" bw list items --search "$name" 2>/dev/null | \
+        python3 -c "import sys,json; items=[i for i in json.load(sys.stdin) if i.get('name','')=='$name']; print(items[0]['id'] if items else '')" 2>/dev/null || true)
+
+    if [ -n "$existing_id" ]; then
+        echo "   Using existing item: $name ($existing_id)"
+        echo "$existing_id"
+        return
```
domenkozar: **P1 — Do not reuse real vault items as mutable fixtures** — When the real vault already contains a generic item such as `Test Database`, this branch reuses it, and Test 22 later replaces its password with `updated-password`. Reused IDs are not recorded in `CREATED_ITEM_IDS`, so cleanup neither deletes nor restores the item and the advertised real-vault suite can corrupt an existing credential.

tests/bitwarden_integration.sh
```diff
@@ -0,0 +1,519 @@
...
+roundtrip_type identity identity
+roundtrip_type ssh ssh
+
+if [ $TESTS_FAILED -eq 0 ]; then
+    echo -e "\n${GREEN}🎉 ALL TESTS PASSED!${NC}"
+    echo "The Bitwarden provider is working correctly with real vault data."
+else
+    echo -e "\n${RED}❌ SOME TESTS FAILED${NC}"
+    echo "Please review the failed tests above."
+fi
```
domenkozar, line 586: **P2 — Exit nonzero when integration assertions fail** — When any `run_test` increments `TESTS_FAILED`, this branch only prints a message and the final `echo` still makes the script exit successfully. Consequently `vaultwarden_harness.sh` treats a failing integration suite as green, allowing provider regressions to pass the documented harness.

domenkozar (commented):
Eight additional findings validated against the current PR head.

domenkozar:
@ashebanow 0.18 targeting Monday, let me know when it's ready for review.

ashebanow:
changes done, just reviewing and testing. Should be done in a few hours.

ashebanow:
I' ve checked in my changes, so the PR is updated, but I haven't had time
to write notes on the PR or your specific comments yet. But you can start
reviewing now if you'd lke.

bsorescu:
Independent verification of the round-2 fixes, on `433aa482` against a disposable Vaultwarden (macOS, bw CLI 2026.6.0, current `vaultwarden/server:latest`), via `RUN_REGRESSIONS=1 tests/vaultwarden_harness.sh`:

- **Regressions: 11/12 FIXED** — R1–R10 and R12 all green, including the substring read/write findings, type-narrowed addressing, the secure-note fallback, built-in fields on create, unknown query-key rejection, and the fixture-adoption fix.
- **Integration suite: all tests passed** (40 assertions) · **collection addressing: 11/11** — and the new config isolation means the harness now survives a developer's `~/.config/secretspec/config.toml`, confirmed with one present.

One finding still reproduces:

```
REPRODUCED R11 — a non-ASCII name is unreachable in lower case
  (got 'Error: × Failed to get secret ╰─▶ Secret 'regr_umlaut' not found')
```

An item created as `Überblick` isn't found when addressed as `überblick`. Since `find_addressed_item` now folds with `to_lowercase`, my read is that the candidate never reaches it: the `bw list items --search <name>` pre-filter (bw.rs:1442) does its own matching, and the lowercase query returns no candidates for the umlaut name. If that's right, a fallback that re-lists without `--search` (or post-filters the full list) when the narrowed query comes back empty would close it — though as a P2 it may also be fine to document and ship.

Everything else looks release-ready from here. I'm closing ashebanow/secretspec#4 — all three of its changes (fail-closed integration exit, XDG_CONFIG_HOME isolation, round-2 regressions) landed natively in this branch in more complete form.

ashebanow:
Well, R11 passes on my machine. But your failure has shown me the test itself is weak, because it depends on 'bw' cli behavior which is version dependent. They've changed behavior fairly recently to strip diacriticals in search (which I'm not convinced is a good way to do this). I'm going to fix up the test to be a little more robust to this.

ashebanow:
Also, I should mention that the right way to do this is by using unicode
normalization and compare on the search terms. I've not done this here
because it would add a unicode dependency to the secretspec binary, and I
didn't want to do that without a go ahead from Domen. It also won't agree
perfectly with the 'bw' cli, bit what its doing is sloppy anyhow.

secretspec/src/provider/bw.rs
ashebanow: Fixed. We are now much more strict about these matches. In fact I've gone beyond what's specified in these comments (here and below) and made it a failure to have an unmatched key at all.

secretspec/src/provider/bw.rs
ashebanow: Fixed. See comment above.

secretspec/src/provider/bw.rs
ashebanow: Fixed.

secretspec/src/provider/bw.rs
ashebanow: Fixed

tests/bitwarden_integration.sh
ashebanow: Fixed.

secretspec/src/provider/bw.rs
ashebanow: Fixed.

secretspec/src/provider/bw.rs
ashebanow: Fixed.

tests/bitwarden_integration.sh
ashebanow: Fixed.

ashebanow:
All changes in now, and individual comments replied to. Thanks to you both
again for all your help.

bsorescu:
Final verification on `c6c32d9`: **full harness green, exit 0** — regressions **12/12 FIXED**, integration suite all passing, collection addressing 11/11.

Worth noting for R11: my runs use bw CLI **2026.6.0**, which predates the `searchCiphersBasic` diacritic fix — so this run exercises the new unfiltered-re-list fallback (`2c90c64`) on exactly the client generation it was written for, and it holds. Nice touch documenting the CLI version boundary in the harness header.

From where I sit this is ready for re-review. @domenkozar

ashebanow:
So I added code coverage support for the bitwarden provider. We were at 75% unit test coverage by line in bw.rs, and 86% line coverage when running all bitwarden related tests (unit, integration, regression, etc). I then used the results to add additional tests targeted at the pieces with no coverage. Now we have:

| scenario | file  | functions | lines  | regions |
|----------|-------|-----------|--------|---------|
| unit     | bw.rs |    97.80% | 95.50% | 94.35%  |
| unit+all | bw.rs |    97.80% | 95.79% | 95.25%  |

The only things that aren't unit tested now are 8 IO closures and 2 test helpers.

ashebanow:
FYI my test shim broke the windows ci, I'm fixing now.

PR #185, opened 2026-07-21
dotenv: reject variable names the parser cannot read back

domenkozar:
## Summary

The dotenv provider wrote any `ref` item verbatim while reading with dotenvy's grammar (`[A-Za-z_][A-Za-z0-9_.]*`). One accepted write of an unrepresentable name, for example an item containing a dash, poisoned the whole file: every later read or write of **any** secret in that store failed to parse at the offending line, and the error surfaced far from the cause.

Found in the wild through the cachix integration (cachix/cachix#737), which built ref items from cache names like `cache-a`.

## Fix

Validate the name against dotenvy's exact grammar in `set`, `get`, and a new `check_writable` override (so the CLI refuses before prompting for a value, matching the env provider's pattern). The rejection names the offending item and how to fix it, consistent with how unsupported `ref` coordinates are rejected.

Interior dots stay accepted deliberately: dotenvy parses them back, and the lenient JS dotenv family (`[\w.-]+`) writes such files, so rejecting them on `get` would break reading legitimate foreign files. Dashes remain impossible: dotenvy can never read them, so accepting the write can only corrupt the store.

## Tests

- `provider/tests.rs` gains a write/read symmetry contract, run against the dotenv and mock providers: a `set` that reports success must be readable back by `get`, and any write, accepted or rejected, must leave previously stored secrets intact.
- A dotenv regression test covers the rejected shapes across `set`/`get`/`check_writable`, that the store survives, and that the sanitized shape (`CACHIX_SIGNING_KEY_CACHE_A`) plus dotted names keep round-tripping.

Verified end to end with the CLI: setting an unrelated secret works, setting the dashed ref item fails cleanly naming the item, and the `.env` remains parseable with the unrelated secret still resolving. All 542 crate tests pass.


domenkozar:

PR #187, opened 2026-07-21
feat(openbao): add first-class OpenBao provider

### LINKED ISSUE #186: Add a first-class OpenBao provider
First, thanks for the work on SecretSpec it’s a thoughtfully designed project, and I’m excited to contribute. I spend quite a bit of time in the OpenBao ecosystem, so the existing support caught my attention.

SecretSpec currently accepts `openbao://` URLs through the Vault provider. This provides basic KV compatibility, but OpenBao is still identified as Vault, URLs are serialized as `vault://`, and configuration follows Vault-specific conventions such as `VAULT_*.`

Would you be open to introducing a dedicated OpenBaoProvider that:

- Registers `openbao://` as its own provider
- Uses OpenBao conventions such as `BAO_*`
- Shares compatible KV and HTTP internals with the Vault provider
- Has its own tests and documentation

The initial PR would focus on first-class provider identity and configuration, not new authentication mechanisms. The separation would also make later OpenBao-specific features possible without adding product-specific branches throughout the Vault provider.

One open question: the vault Cargo feature currently enables both schemes. Would you prefer OpenBao to remain under that feature for compatibility, or receive a separate openbao feature?

I’d be happy to work on the implementation once the preferred direction is clear.

dc-tec:
## Summary

Add first-class OpenBao support with its own `openbao` provider identity, URI scheme, Cargo feature, configuration conventions, documentation, and integration-test registration.

The Vault and OpenBao providers remain separate publicly while sharing their compatible KV and authentication implementation internally.

Closes #186.

## What changed

- Register `openbao://` as an independent provider rather than an alias of Vault.
- Add a standalone `openbao` Cargo feature and include it in the default feature set.
- Preserve support for KV v1 and KV v2, namespaces, token authentication, AppRole, and JWT/OIDC.
- Use OpenBao configuration variables:
  - `BAO_ADDR`
  - `BAO_NAMESPACE`
  - `BAO_TOKEN`
  - `BAO_TOKEN_PATH`
- Add OpenBao-prefixed SecretSpec inputs for AppRole and JWT authentication, with corresponding `VAULT_*` variables retained as compatibility fallbacks.
- Add OpenBao to the generic provider integration suite and standalone feature checks.
- Add a dedicated OpenBao provider guide and update every provider listing, example, and reference location with the upcoming `0.17+` compatibility boundary.
- Refocus the existing Vault documentation on HashiCorp Vault while documenting the practical OpenBao fallback available in SecretSpec 0.16.

## Resolved findings

While extracting the shared Vault-compatible client, this PR also fixes two pre-existing issues:

- Addresses ending in `/` could produce invalid `//v1/...` API paths. Trailing separators are now normalized before request paths are appended.
- AppRole and JWT login requests did not include `X-Vault-Namespace`, causing namespaced authentication to fail. Login and authenticated data requests now share the namespace-header construction.

Both fixes apply to Vault and OpenBao and include regression tests.

## Compatibility

This does not introduce new authentication protocols or change the compatible wire API. OpenBao intentionally retains the `X-Vault-*` protocol headers.

Default builds include both providers. Builds selecting provider features explicitly must enable `openbao` to use `openbao://` starting with SecretSpec 0.17.

## Validation

- OpenBao 2.6.0 integration testing:
  - Generic provider workflow against KV v1 and KV v2
  - Token-authenticated CLI set/get with a trailing-slash `BAO_ADDR`
  - Namespaced AppRole set/get round trip

### DIFF
--- a/docs/src/content/docs/providers/openbao.md
+++ b/docs/src/content/docs/providers/openbao.md
@@ -209,7 +209,7 @@ $ secretspec check --provider openbao://bao.example.com:8200/secret
 ### Development mode
 
 ```bash
-$ bao server -dev
-$ export BAO_TOKEN=hvs.dev-root-token
+$ bao server -dev -dev-root-token-id="dev-only-token"
+$ export BAO_TOKEN="dev-only-token"
 $ secretspec check --provider "openbao://127.0.0.1:8200/secret?tls=false"
 ```
--- a/docs/src/content/docs/providers/openbao.md
+++ b/docs/src/content/docs/providers/openbao.md
@@ -1,0 +1,215 @@
+---
+title: OpenBao Provider
+description: OpenBao integration, available in SecretSpec 0.17+
+---
+
+The OpenBao provider integrates with OpenBao's KV (Key-Value) secrets engine
+using OpenBao's own provider identity and configuration conventions.
+
+:::caution[Version compatibility]
+The `openbao` provider targets SecretSpec 0.17 and is unavailable
+in the current 0.16 release. SecretSpec 0.16 can still connect to OpenBao with
+an `openbao://` URI through the `vault` build feature, but reports the provider
+as Vault and uses `VAULT_*` environment variables.
+:::
+
+## Using SecretSpec 0.16 today
+
+The current release's practical form is:
+
+```bash
+$ export VAULT_TOKEN=hvs.your-token-here
+$ secretspec check --provider openbao://bao.example.com:8200/secret
+```
+
+For a minimal Rust build, enable `vault`:
+
+```toml
+secretspec = { version = "0.16", default-features = false, features = ["vault"] }
+```
+
+## At a glance
+
+| | |
+| --- | --- |
+| Provider | `openbao` (0.17+) |
+| URI | `openbao://[namespace@]host[:port][/mount][?options]` |
+| Access | Read and write; secret references are read-only |
+| Best for | Open-source, policy-controlled secret infrastructure |
+| Authentication | Token, AppRole, or JWT/OIDC |
+| Build feature | `openbao` (0.17+) |
+| Default storage | KV path `secretspec/{project}/{profile}/{key}`, field `value` |
+
+## Quick start
+
+```bash
+$ export BAO_TOKEN=hvs.your-token-here
+$ secretspec set DATABASE_URL --provider openbao://bao.example.com:8200
+Enter value for DATABASE_URL: postgresql://localhost/mydb
+✓ Secret 'DATABASE_URL' saved to openbao (profile: default)
+```
+
+## Setup
+
+### Prerequisites
+
+- A running OpenBao server
+- Authentication credentials
+- KV secrets engine enabled (v1 or v2)
+- Build with `--features openbao`
+
+### Environment compatibility
+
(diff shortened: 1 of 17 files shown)


secretspec/src/provider/vault_common.rs
```diff
@@ -0,0 +1,895 @@
...
+    ///
+    /// This retains the historical compact form while using the actual product
+    /// scheme, which is the identity bug the first-class split fixes.
+    pub(crate) fn uri(&self) -> String {
+        let host = self
+            .config
+            .endpoint
+            .trim_start_matches("https://")
+            .trim_start_matches("http://");
+        let mut uri = format!("{}://{host}", self.product.scheme());
```
domenkozar: **[P1] Sanitize environment-derived endpoints before reporting them.** When `endpoint` comes from `BAO_ADDR`/`VAULT_ADDR`, this string stripping preserves userinfo and query data. For example, `BAO_ADDR=https://alice:secret@bao:8200` makes `uri()` return `openbao://alice:secret@bao:8200`. Callers treat `Provider::uri()` as credential-free and write it to audit records, warnings, and reports without another redaction pass, so this can disclose credentials. Please parse and reconstruct the endpoint without password/userinfo or sensitive query parameters.

secretspec/src/provider/vault_common.rs
```diff
@@ -0,0 +1,895 @@
...
+        if self.config.mount != "secret" {
+            uri.push('/');
+            uri.push_str(&self.config.mount);
+        }
+        uri
```
domenkozar: **[P2] Preserve the provider's non-secret configuration in `uri()`.** This reconstruction drops the namespace plus `tls`, KV version, auth method, role, and audience. Thus `openbao://team-a@bao:8200/secret?tls=false&kv=1` is reported as `openbao://bao:8200`, making provenance and audit output unable to distinguish materially different stores. The `Provider::uri()` contract says the full non-secret attribution—including namespace and path—should be retained.

docs/src/content/docs/providers/openbao.md
```diff
@@ -0,0 +1,215 @@
...
+
+$ export BAO_NAMESPACE=team-a
+$ secretspec check --provider openbao://bao.example.com:8200/secret
+```
+
+### Development mode
+
+```bash
+$ bao server -dev
+$ export BAO_TOKEN=hvs.dev-root-token
```
domenkozar: **[P2] Use the token configured for this dev server.** `bao server -dev` does not select the literal `hvs.dev-root-token`, so the next command will authenticate with an unrelated token and fail. Please either tell the reader to copy the token printed by the server, or use `bao server -dev -dev-root-token-id=dev-only-token` and export `BAO_TOKEN=dev-only-token`, as in the [official OpenBao quick start](https://openbao.org/docs/get-started/developer-qs/).

domenkozar (commented):
I found three actionable issues in the new provider URI reporting and OpenBao development documentation.

dc-tec:
> I found three actionable issues in the new provider URI reporting and OpenBao development documentation.

Thanks for the review, all three findings have been addressed. 

Regression tests cover endpoint sanitization and preservation of non-secret provider attribution. 

The OpenBao development example now uses an explicit -dev-root-token-id

PR #193, opened 2026-07-23
feat: secret scopes (--scope), membership-only subsets of a profile

ojhermann:
@domenkozar this re-establishes #163, which was closed automatically when I deleted my fork.


secretspec/src/secrets.rs
```diff
@@ -3304,14 +3621,35 @@ impl Secrets {
...
+            || plan.secrets.len() == compiled_profile.secrets.len())
+        .then_some(&compiled_profile.constraints);
         let mut constraint_violations = Vec::new();
         if let Some(constraints) = constraints {
+            // Under a scope a group is judged on the members the consumer can
+            // actually see: a group with no visible member is not this
+            // consumer's concern, and one with some is enforced over those. The
+            // reported member list is narrowed the same way, so a violation
+            // message never names a secret the scope hides.
+            let visible_members = |members: &Vec<String>| -> Vec<String> {
```
domenkozar, line 3710: A scoped constraint can be narrowed to a single visible member here (the new `aws` test even asserts `violation.secrets == ["AWS_KEY"]`), but `schema/resolution-report.schema.json` still requires `constraintViolation.secrets` to have `minItems: 2` and describes it as all declared members. Consequently, `check --scope ... --json` can emit a payload that fails the canonical SecretSpec schema. Please update the schema for the scoped representation and add coverage that validates a serialized scoped violation against it.

secretspec/src/secrets.rs
```diff
@@ -1199,7 +1351,59 @@ impl Secrets {
...
+        // An active scope narrows the profile to the intersection with its
+        // membership. This is the single upstream worklist for validation,
+        // resolution, planning, prompting, generation, and audit attribution, so
+        // filtering here scopes every one of them at once — and an unknown scope
+        // fails before any provider is touched.
+        let members = self.active_scope_members()?;
+        Ok(compiled
+            .secrets
+            .keys()
+            .filter(|name| members.as_ref().is_none_or(|m| m.contains(name.as_str())))
```
domenkozar: This helper is also used by `import()` to build its copy worklist. Because `active_scope_members()` reads ambient `SECRETSPEC_SCOPE`, `secretspec import` can now silently copy only the scoped subset even though `import` has no `--scope` option and the documented scope surface is `check`/`run`/`export`. This is especially easy to trigger from a shell or child process that inherited the variable. Please keep import on an unscoped profile-name path, or make scope applicability explicit at the caller.

ojhermann:
Looking into https://github.com/cachix/secretspec/actions/runs/30024972250/job/89270518051

ojhermann:
Fixed in 9057a03. The four failing `scopes` tests drove a child via `sh -c 'printf ... > $path'` with a Windows temp path, whose backslashes get mangled inside `sh` so the file was never written — gated them to `#[cfg(unix)]`, matching the existing `test_run_cleans_up_as_path_temp_files`.

domenkozar:
Follow-up implemented in `f6148b7`:

- The shared JSON/FFI request now accepts `scope`, and both resolve and report responses return the active scope.
- The Go, Node.js, Python, Ruby, PHP, C#, and Haskell SDK builders expose explicit scope selection and surface the returned scope.
- Audit events for scoped `check`, `run`, and `export` now include the scope name. Scoped `check` records the keys actually accessed (including hidden composition dependencies), while `run` and `export` record the visible/exposed keys.
- Added FFI, SDK, core audit, and CLI coverage, plus 0.17+ SDK/audit documentation and changelog updates.

The branch is rebased onto the latest `origin/main` (`f4f333d`). Validation passed: `cargo test --all` (635 core tests), the full SDK matrix, Clippy with warnings denied, rustfmt, and the docs build.

PR #196, opened 2026-07-24
Add Scaleway Secret Manager provider

cramt:
Adds a `scaleway://` provider backed by [Scaleway Secret Manager](https://www.scaleway.com/en/secret-manager/) via its `v1beta1` REST API. Gated behind a `scaleway` build feature (on by default).

## Behavior

- **Auth:** API secret key in the `X-Auth-Token` header, from the `secret_key` credential or `SCW_SECRET_KEY`.
- **Targeting:** region from the URI host / `SCW_DEFAULT_REGION` / `fr-par`; project from `?project_id=` / `SCW_DEFAULT_PROJECT_ID`.
- **Storage:** Scaleway secret names can't contain `/`, so the SecretSpec convention lives in the folder hierarchy: convention secrets are stored at path `[{base}/]secretspec/{project}/{profile}` with the key as the secret name. `set` creates the secret if needed, then appends a version; `get` reads the latest enabled version and base64-decodes it.
- **Native refs:** `item` is an absolute Scaleway path; optional `field` extracts one key of a JSON (`key_value`) secret and `version` pins a revision. Native refs are read-only (a single-field write would append a version discarding sibling keys).

URI: `scaleway://[REGION][?project_id=UUID&path=/folder]`

## Docs

Updated every listing in the provider checklist: dedicated provider page, concepts + reference tables, security-considerations table, sidebar + LLM summary, landing-page grid, quick-start output, and both READMEs. New provider is version-gated as `(0.17+)`.

## Testing

- Unit tests for URI/config parsing, path normalization, convention/native addressing, `uri()` round-trip, JSON-key extraction, base64 decode, read-only refs, and registry resolution.
- Verified end-to-end against the live Scaleway API: `set` → `get` round-trips the exact value (exercising create-secret → add-version → access-by-path → base64 decode); a bogus key surfaces a clean `401`. Test secrets were created under a throwaway folder and deleted afterward.

### DIFF
--- a/docs/src/pages/index.astro
+++ b/docs/src/pages/index.astro
@@ -43,6 +43,7 @@ const providerMetadata: ProviderMetadata[] = [
   { icon: 'vault', slug: 'vault', label: 'Vault' },
   { slug: 'openbao', label: 'OpenBao (0.17+)' },
   { slug: 'awssm', label: 'AWS Secrets Manager' },
+  { slug: 'scaleway', label: 'Scaleway Secret Manager (0.17+)' },
   { icon: 'googlecloud', slug: 'gcsm', label: 'Google Cloud Secret Manager' },
   { icon: 'microsoftazure', slug: 'akv', label: 'Azure Key Vault' },
   { slug: 'infisical', label: 'Infisical' },
--- a/docs/src/pages/index.astro
+++ b/docs/src/pages/index.astro
@@ -150,6 +150,7 @@ $ secretspec config global init
   openbao: OpenBao secret management (0.17+)
   age: age-encrypted file (0.17+)
   systemd-credential: Read-only systemd service credentials (0.17+)
+  scaleway: Scaleway Secret Manager (0.17+)
 <span class="tok-str">✓</span> Saved to ~/.config/secretspec/config.toml
 
 <span class="tok-com"># 3. Run your app with secrets injected</span>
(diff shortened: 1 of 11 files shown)


secretspec/src/provider/scaleway.rs
```diff
@@ -0,0 +1,721 @@
...
+
+        let region = url
+            .host()
+            .filter(|s| !s.is_empty())
+            .or_else(|| preferred_env(&[REGION_ENV]))
+            .unwrap_or_else(|| DEFAULT_REGION.to_string());
+
+        let project_id = url.query_value("project_id").filter(|s| !s.is_empty());
+
+        let path = normalize_path(url.query_value("path").as_deref().unwrap_or("/"));
```
domenkozar, line 107: Please reject non-root URI paths here. `scaleway://fr-par/team` is parseable, but `url.path()` is currently ignored and `uri()` reconstructs the provider without `/team`, so reads or writes silently fall back to the root `secretspec/...` hierarchy rather than the intended namespace. Since folders are configured through `?path=...`, this malformed form should fail like it does for the other cloud providers. Please add a regression test for it as well.

docs/src/pages/index.astro
```diff
@@ -43,6 +43,7 @@ const providerMetadata: ProviderMetadata[] = [
   { icon: 'vault', slug: 'vault', label: 'Vault' },
   { slug: 'openbao', label: 'OpenBao (0.17+)' },
   { slug: 'awssm', label: 'AWS Secrets Manager' },
+  { slug: 'scaleway', label: 'Scaleway Secret Manager (0.17+)' },
```
domenkozar, line 46: Please also add `scaleway: Scaleway Secret Manager (0.17+)` to both `secretspec config init` output samples on this page (around lines 145 and 216) and to the full selector output in `README.md`. The provider registry now includes Scaleway, so these three examples should match the CLI output.

domenkozar:
I'm planning to release 0.17 on Monday so if you get changes in we can merge it! :)

cramt:
> I'm planning to release 0.17 on Monday so if you get changes in we can merge it! :)

That'd be great, cause we're looking to adopt it at my company, and we just need this

PR #205, opened 2026-07-27
feat(providers): general ?layout=flat setting to resolve at the store root

### LINKED ISSUE #181: Infisical provider: `?layout=flat` opt-in to resolve from the environment root (no `{project}/{profile}` folders)
Tracking issue for the flat-layout option discussed on #154. @domenkozar gave the go-ahead to explore it ([#154 comment](https://github.com/cachix/secretspec/pull/154#issuecomment-5026278021)); raised by @nickdichev migrating from BWS.

## Motivation

The Infisical provider namespaces every secret under `{path}/{project}/{profile}` (default `path` = `/secretspec`), in the environment named by the profile. That default keeps multiple projects/profiles from colliding and preserves profile separation even when `?env=` is pinned. But for a **single-project store** — e.g. a BWS→Infisical migration where secrets already live at the environment root — the scaffolding is surprising: root secrets simply don't resolve, and the current workarounds are Infisical *imports* or a per-secret leading-slash `ref` (`{ ref = { item = "/DATABASE_URL" } }`).

## Proposal

A provider query opt-in, `?layout=flat`, that resolves (and writes) secrets at the **environment root** instead of under `{project}/{profile}`. Default stays the current namespaced layout — no behavior change for existing configs.

## Open design points to settle in the investigation

1. **Profile separation.** Flat + a pinned `?env=` collapses every profile onto the same key (the profile folder is what separates them today). This should be a deliberate, documented "you're giving up profile separation" mode — decide whether flat *requires* `?env=`, or whether the profile still names the environment (just not a folder).
2. **Interaction with `?path=`.** Is `flat` "no `{project}/{profile}`, but still honor `path`", or truly the environment root? (Likely: flat drops the `{project}/{profile}` segments but a `?path=` prefix, if given, still applies.)
3. **Writes.** `set` under flat writes to the root too — confirm symmetry and that `check_writable` is unaffected.
4. **Spelling.** `?layout=flat` (vs. `nested`/default) — is `layout` the right knob name, and do we want to name the default explicitly?

## Plan

Investigation + design here, then a resolving PR for @domenkozar to review. Docs updated per the provider-doc checklist; `(0.16+)`/target-version labeling since the Infisical provider is still unreleased.

Raised in #154.

ojhermann:
Resolves #181. Reopens #183, which I closed by accident in a branch sweep — rebased onto `main`, with Scaleway added.

This started as an Infisical-only `?layout=flat` (raised by @nickdichev migrating from BWS); per your [suggestion](https://github.com/cachix/secretspec/issues/181#issuecomment-5026947869) it is a **general provider setting**, since the namespaced-vs-root question is the same for every hierarchical backend.

## What this adds

`?layout=flat` maps SecretSpec's `{project}/{profile}/{key}` convention onto the **store root** — the secret is addressed by key alone, dropping the `secretspec/{project}/{profile}` scaffolding but keeping any provider-native container (a Vault mount) or user prefix (an AWS `?prefix=`, an Infisical `?path=`). The default nested layout is unchanged, so existing configs behave exactly as before.

Honored by **Infisical, Vault, OpenBao, AWS Secrets Manager, Google Cloud Secret Manager, Azure Key Vault and Scaleway Secret Manager**. Providers with no hierarchy (`dotenv`, `env`, `bws`) are already flat and ignore it.

```toml title="secretspec.toml"
[providers]
vault  = "vault://vault.example.com:8200/secret?layout=flat"
awssm  = "awssm://us-east-1?prefix=myteam&layout=flat"   # -> myteam/DATABASE_URL
```

Two caveats are documented at each point of use, since both give something up:

- **Profile separation.** Dropping the profile means distinct profiles no longer separate themselves by name. Combining flat with a provider that pins its environment another way (an Infisical `?env=`) collapses every profile onto one key — deliberate, but only safe when a single profile is ever resolved against that store.
- **The key must fit the store.** With no `secretspec-` rewriting to fall back on, the key itself must be a legal name in the backend. Azure Key Vault is strictest: under flat the key is used verbatim, so an underscore — which the nested layout would have Base32-encoded away — is refused.

An unreadable `?layout=` value is refused rather than guessed.

## Since #183 was closed

Scaleway landed on `main` in the meantime and is hierarchical in the same way Infisical is, so it honors the setting too (second commit, kept separate for review). Without it, "every hierarchical provider" would not have been true.

## Still open, and yours to call

The question from #183 that never got settled: **`?layout=flat` as a query parameter, or a config field?** This PR implements the query-parameter form on the reasoning that providers are URI-configured today (`env`/`path`/`tls` already are) and aliases are defined once, so it is set once anyway. Say the word if you would rather it were a field and I will move it — that is a cheap change while this is still unreleased, and expensive afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Injected credentials read through a private, owner-only `dcli` state
   directory of their own, because `dcli` otherwise prefers a device already
   registered on the machine and reads that identity's vault instead.
+- Flat layout (`?layout=flat`), a general provider setting that resolves and
+  writes convention secrets by key alone at the store root, dropping the
+  `secretspec/{project}/{profile}` scaffolding — the natural shape for a
+  single-project store, such as one migrated from another secret manager.
+  Supported by Infisical, Vault, OpenBao, AWS Secrets Manager, Google Cloud
+  Secret Manager, Azure Key Vault and Scaleway Secret Manager; providers that
+  address by key alone (`dotenv`, `env`, `bws`, `sops`) are already flat and
+  ignore it. Any
+  provider-native container (a Vault mount) or user-supplied prefix (an AWS
+  `?prefix=`, an Infisical `?path=`) still applies. The default nested layout is
+  unchanged. Dropping the profile gives up profile separation, so a provider
+  that pins its environment another way (an Infisical `?env=`) combined with
+  `?layout=flat` collapses every profile onto one key deliberately; and because
+  the key is used as the backend name directly, it must be a legal name there
+  (Azure Key Vault, which uses it verbatim, refuses a key it cannot store).
 
 ## [0.17.0] - 2026-07-26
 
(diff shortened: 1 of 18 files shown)


ojhermann:
Moved to draft; see https://github.com/cachix/secretspec/issues/181#issuecomment-5125863617

ojhermann:
Superseded by the trait-level naming template discussed in https://github.com/cachix/secretspec/issues/181#issuecomment-5169000791 — flat becomes a `{key}` template rather than its own setting.

PR #216, opened 2026-07-29
feat: add Dashlane provider (dashlane://)

### LINKED ISSUE #128: Support Dashlane as a provider
Request to support [Dashlane](https://www.dashlane.com/) as a provider.

https://cli.dashlane.com/

ojhermann:
Closes https://github.com/cachix/secretspec/issues/128

Adds a `dashlane://` provider backed by the [Dashlane CLI](https://cli.dashlane.com/). `dcli` **cannot write** — no `create`/`add`/`set`/`update`/`delete` for any item type ([their docs](https://cli.dashlane.com/personal/vault): "For now accessing your vault is read only"; maintainer confirmation in https://github.com/Dashlane/dashlane-cli/issues/212). So it is read-only like `env`: `check_writable` states that reason and `set` returns it before prompting for a value.

Convention secrets read the item titled `secretspec/{project}/{profile}/{key}`; a `ref` names an existing item by title or identifier with an optional `field`. `dashlane://note|secret|password` pins one content type, plain `dashlane://` searches all three. Auth is `dcli sync` device registration, or `DASHLANE_SERVICE_DEVICE_KEYS` for CI, also injectable as the `service_device_keys` credential.

Reads go through the `-o json` listers, not `dcli read`: on a duplicate title `read` returns `rc=0` and one **arbitrary** item, and its path parser rejects both `/` in a title and the braced id form `dcli` itself prints. Matching is exact here and an ambiguous title is refused. That also gives `get_many` one listing per content type instead of one subprocess per secret.

Marked `(0.18+)`, since 0.17.0 is released.

Partly verified against dcli 6.2628.1 on Linux: `dcli status` on an unregistered device prints `Logged in: no` to stdout and exits 0, and a lister with stdin closed exits 1 with an ANSI-coloured `error: User force closed the prompt with 0 null` — so the closed stdin does turn an auth prompt into a clean failure rather than a hang. That last string is why the auth-error match is what it is; my first guesses (`not logged in`, `No device configuration`) appear nowhere in it.

What is still unverified is everything needing an actual vault: the item JSON deserializing, a successful read, and the `Logged in: yes` / `Locked:` lines. The `secret` content type is the weakest of those — Dashlane Secrets are a Business-plan feature, so its `content` field mapping comes from `VaultSecret` in dcli's own types rather than an observed vault, and no personal account can exercise it. There are `#[ignore]`d live tests covering exactly that for whoever has one: `cargo test -p secretspec provider::dashlane::live -- --ignored --nocapture`.

### DIFF
--- a/docs/src/content/docs/providers/dashlane.md
+++ b/docs/src/content/docs/providers/dashlane.md
@@ -22,6 +22,7 @@ here.
 | Access | Read-only |
 | Best for | Teams already keeping developer secrets in Dashlane |
 | Authentication | `dcli` device registration, or `DASHLANE_SERVICE_DEVICE_KEYS` |
+| Network | Local reads; `dcli` re-syncs when its copy is over an hour old |
 | Default storage | Item titled `secretspec/{project}/{profile}/{key}` |
 
 ## Quick start
@@ -80,7 +81,9 @@ dashlane://[ITEM_TYPE]
 
 `ITEM_TYPE` restricts the search to one Dashlane content type: `secret`,
 `note`, or `password` (`login` is accepted as an alias for `password`). Omit it
-to search secrets, then notes, then logins — the same order `dcli read` uses.
+to search secrets, then logins, then notes — the order `dcli read` resolves a
+name in, so a title held by both a login and a note resolves the same way here
+as it does through `dcli`.
 
 Pinning the type is worth doing when you know where your secrets live: each
 content type searched costs one `dcli` invocation.
@@ -175,10 +178,12 @@ every login and SSO are not supported for them.
 `dcli sync`, then read it here. Give the secret a writable provider as well if
 you need to set values from SecretSpec.
 
-**A new item is invisible until the CLI syncs.** `dcli` reads a locally
-synced copy of the vault and re-syncs hourly. SecretSpec never syncs on your
-behalf — a sync is a network round-trip, and a secret read should not silently
-become one. Run `dcli sync` after adding an item.
+**A new item is invisible until the CLI syncs.** `dcli` reads a local copy of
+the vault. SecretSpec never asks it to sync, but `dcli` syncs itself whenever
+its last sync is over an hour old, so a read is usually local and occasionally
+a network round-trip. Run `dcli sync` after adding an item rather than waiting
+for that, and `dcli configure disable-auto-sync true` to keep reads strictly
+offline.
 
 **Only three content types are readable.** `dcli` exposes listers for secrets,
 secure notes, and logins. Passkeys, personal info, payments, and IDs have no
@@ -190,7 +195,8 @@ personal plans, where `dashlane://secret` finds nothing. Use secure notes there.
 ## Security considerations
 
 Reads decrypt the local vault in a `dcli` subprocess and pass the value back
-over a pipe; SecretSpec never writes it to disk or to a log. `dcli password`
+over a pipe; SecretSpec never writes it to disk or to a log. A read can reach
+the network, because `dcli` re-syncs when its copy is over an hour stale. `dcli password`
 copies a password to the system clipboard when it is run without an output
 format, so SecretSpec always requests JSON explicitly.
 
--- a/docs/src/content/docs/providers/dashlane.md
+++ b/docs/src/content/docs/providers/dashlane.md
@@ -65,10 +65,10 @@ By default the master password is saved in the OS keychain. `dcli lock` locks
 the vault again, and `dcli logout` clears both the local database and the
 keychain entry.
 
-SecretSpec checks `dcli status` before reading and fails with instructions if
-the device is unregistered or the vault is locked. It never answers a `dcli`
-prompt: reads run with stdin closed, so an unauthenticated CLI fails
-immediately instead of hanging a `secretspec run`.
+SecretSpec checks `dcli status` before reading and reports an unregistered
+device or a locked vault. It never answers a `dcli` prompt — reads run with
+stdin closed, so an unauthenticated CLI fails immediately rather than leaving
+`secretspec run` waiting on a password.
 
 ## Configuration
 
@@ -126,22 +126,21 @@ To read an item you already have, name it with a `ref`:
 ```toml title="secretspec.toml"
 [profiles.default]
 # By title
-GITHUB_TOKEN = { ref = { item = "GitHub personal access token" } }
+GITHUB_TOKEN = { description = "GitHub token", ref = { item = "GitHub personal access token" }, providers = ["dashlane"] }
 
-# By identifier — stable across renames, and unambiguous
-STRIPE_KEY = { ref = { item = "D47734C4-0ABE-423A-8633-6B9F10A38905" } }
+# By identifier — stable across renames, and never ambiguous
+STRIPE_KEY = { description = "Stripe key", ref = { item = "D47734C4-0ABE-423A-8633-6B9F10A38905" }, providers = ["dashlane"] }
 
 # A named field of a login
-DB_USER = { ref = { item = "Production database", field = "login" } }
+DB_USER = { description = "Database user", ref = { item = "Production database", field = "login" }, providers = ["dashlane://password"] }
 ```
 
-Find an item's identifier with `dcli password <title> -o json`; the `id` field
-is emitted wrapped in braces, and either form works here.
+Find an item's identifier by listing its content type — `dcli secret`,
+`dcli note`, or `dcli password` with `-o json`. The `id` is emitted wrapped in
+braces; both forms work here.
 
 `ref` supports the `item` and `field` coordinates. Dashlane has no vaults or
 sections, so those coordinates are rejected rather than ignored.
-
-Referenced items are read-only like everything else in this provider.
 
 ## CI/CD
 
@@ -172,10 +171,9 @@ every login and SSO are not supported for them.
 
 ## Troubleshooting and limitations
 
-**Writes are not supported.** `dcli` has no `create`, `add`, `set`, `update`,
-or `delete` subcommand for vault items, so `secretspec set` fails with that
-reason. Add or edit the item in a Dashlane app, run `dcli sync`, then read it
-here. Pair Dashlane with a writable provider if you need `secretspec set`.
+**`secretspec set` fails.** Add or edit the item in a Dashlane app, run
+`dcli sync`, then read it here. Give the secret a writable provider as well if
+you need to set values from SecretSpec.
 
 **A new item is invisible until the CLI syncs.** `dcli` reads a locally
 synced copy of the vault and re-syncs hourly. SecretSpec never syncs on your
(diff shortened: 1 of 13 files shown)


secretspec/src/provider/dashlane.rs
```diff
@@ -0,0 +1,1029 @@
...
+    ///
+    /// Returns bytes rather than a `String`: on the lister path stdout is a
+    /// plaintext dump of vault items, so it must never reach a log or an error
+    /// message.
+    fn run(&self, args: &[&str]) -> Result<Vec<u8>> {
+        let mut cmd = Command::new("dcli");
+        // Injected credentials reach `dcli` the only way it reads them,
+        // without touching this process's own environment.
+        if let Some(keys) = self.device_keys() {
+            cmd.env(DEVICE_KEYS_ENV, keys);
```
domenkozar, line 385: **[P1] Isolate dcli state when injecting device keys**

When `dcli` already has a device database, or two credential-scoped Dashlane aliases run sequentially, upstream `getLocalConfiguration` prefers the existing device row and ignores `DASHLANE_SERVICE_DEVICE_KEYS`. Merely setting the environment variable here can therefore silently read matching items from the previously registered vault instead of the requested identity. Please use credential-scoped `dcli` state via an isolated `HOME`/`APPDATA`, or detect and reject this conflict.

secretspec/src/provider/dashlane.rs
```diff
@@ -0,0 +1,1029 @@
...
+        match self {
+            Self::Secret | Self::Note => "content",
+            Self::Password => "password",
+        }
+    }
+
+    /// The search order used when the URI pins no single type, matching the
+    /// precedence `dcli read` applies across the vault.
+    fn search_order() -> &'static [Self] {
+        &[Self::Secret, Self::Note, Self::Password]
```
domenkozar: **[P2] Match dcli login-before-note precedence**

When a login and secure note have the same title and no item type is pinned, this order returns the note. `dcli read` selects secrets, then credentials/logins, then notes, so the provider returns a different value despite claiming to mirror `dcli`; place `Password` before `Note` or document the intentional divergence.

secretspec/src/provider/dashlane.rs
```diff
@@ -0,0 +1,1029 @@
...
+    /// `dcli` emits one and accepts the other — or the title, compared
+    /// case-insensitively as `dcli read` compares it.
+    fn matches(&self, name: &str) -> bool {
+        let bare = name.trim_start_matches('{').trim_end_matches('}');
+        if !bare.is_empty() && self.bare_id().eq_ignore_ascii_case(bare) {
+            return true;
+        }
+        self.title
+            .as_deref()
+            .is_some_and(|title| title.eq_ignore_ascii_case(name))
```
domenkozar: **[P2] Preserve Unicode case-insensitive title matching**

For a referenced title such as `Überblick` addressed as `überblick`, `eq_ignore_ascii_case` returns false because it only folds ASCII. The Dashlane matcher lowercases Unicode strings and the new documentation promises case-insensitive title matching, so these references incorrectly appear unset or fall through to another provider.

docs/src/content/docs/reference/providers.md
```diff
@@ -397,6 +416,7 @@ export SECRETSPEC_PROVIDER="dotenv:///config/.env"
 | GoPass | ✅ GPG encryption | Local filesystem | ❌ No |
 | Proton Pass | ✅ End-to-end | Cloud (Proton) | ✅ Yes |
 | LastPass | ✅ End-to-end | Cloud (LastPass) | ✅ Yes |
+| Dashlane (0.18+) | ✅ End-to-end | Cloud (Dashlane), synced locally | Only on `dcli sync` |
```
domenkozar: **[P2] Account for automatic hourly dcli sync**

With default `dcli` configuration and a last sync older than one hour, each lister calls upstream `connectAndPrepare({})`, which automatically synchronizes the vault. Consequently `secretspec check` or `run` can perform network access without an explicit `dcli sync`, contrary to this table and the provider page; document the hourly access or require users to disable auto-sync.

ojhermann:
All four addressed in 5f1c21d — thanks, the device-keys one was the serious find.

Confirmed it upstream before fixing: `getEnvDeviceCredentials` is only reached from `getLocalConfigurationWithoutDB`, which `getLocalConfiguration` calls only when there is no device row, so with one present the variable is ignored entirely and the read silently goes to the wrong identity. Each credential now gets its own `HOME`/`APPDATA` under the cache dir, named after a hash of the keys rather than the keys themselves. Search order is now secrets → logins → notes to match `secrets[0] ?? credentials[0] ?? notes[0]`, titles fold with `to_lowercase`, and the sync claims are corrected on both the provider page and the security table — `connectAndPrepare` does sync when the last one is over an hour old, so I've documented that and pointed at `disable-auto-sync` for anyone who wants reads strictly local. Three regression tests cover the order, the non-ASCII fold, and that the state directory never contains the keys.

Separately, the `build` and `sdks` failures on the previous run were a bad binary-cache entry — both died at `Evaluating shell` on `path '/nix/store/i9nx0dp1khrgikqr95ryy2jkigr4c5yv-unpack-bootstrap-tools.sh' is not valid`, before compiling anything, while `tests` on both runners passed. This push should get a clean run.

secretspec/src/provider/dashlane.rs
ojhermann: Fixed in 5f1c21d. Each injected credential now gets its own `HOME`/`APPDATA` under the cache dir, so `dcli` starts from a state directory with no device row and actually reads the requested identity. The directory is named after a hash of the keys rather than the keys, since a path is visible to anyone who can list the cache. A test asserts that, and that the name is stable per credential so the synced vault is reused.

secretspec/src/provider/dashlane.rs
ojhermann: Fixed in 5f1c21d — the order is now secrets → logins → notes. Thanks for catching it; I had the `secrets[0] ?? credentials[0] ?? notes[0]` line in hand earlier and still wrote notes second. A test now pins the relative order, and the provider page and reference docs are updated to match.

secretspec/src/provider/dashlane.rs
ojhermann: Fixed in 5f1c21d. Titles now fold with `to_lowercase`; identifiers keep `eq_ignore_ascii_case`, which is exact for UUIDs. Regression test uses your `Überblick` / `überblick` case.

docs/src/content/docs/reference/providers.md
ojhermann: Fixed in 5f1c21d. You are right that a lister can sync on its own, so I have documented the hourly behaviour rather than requiring users to disable it: the security table now reads "Yes — `dcli` auto-syncs hourly", and the provider page says a read is usually local and occasionally a round-trip, pointing at `dcli configure disable-auto-sync true` for anyone who wants reads strictly offline. The same claim was wrong in the provider struct docs and is corrected there too.

secretspec/src/provider/dashlane.rs
```diff
@@ -0,0 +1,1143 @@
...
+        if let Some(keys) = self.device_keys() {
+            // ...but only where no device is already registered. `dcli` reads
+            // DASHLANE_SERVICE_DEVICE_KEYS inside `getLocalConfigurationWithoutDB`,
+            // which upstream calls only when its state directory holds no device
+            // row; with one present, `getLocalConfiguration` takes that row's
+            // login instead and the variable is ignored. On a machine already
+            // logged in interactively, or across two aliases carrying different
+            // keys, that silently reads the wrong identity's vault. Giving each
+            // credential its own state directory keeps the identities apart.
+            if let Some(dir) = scoped_state_dir(&keys) {
```
domenkozar: **[P1] Fail closed if scoped state cannot be resolved**

`scoped_state_dir` returns `None` when the cache location cannot be resolved, but this `if let` then falls through and still starts `dcli` with the injected keys and inherited state paths. If inherited `APPDATA` points at an existing device database, `dcli` again ignores the keys and reads the wrong identity, which is the original failure mode. Please make scoped-directory resolution return an error and abort before spawning instead of silently running unisolated.

secretspec/src/provider/dashlane.rs
```diff
@@ -0,0 +1,1143 @@
...
+                // `dcli` derives the directory from APPDATA, else HOME, and
+                // creates it recursively itself.
```
domenkozar: **[P1] Create credential-scoped state with private permissions**

Letting `dcli` create this tree does not make it private: its recursive Node `mkdir` uses mode `0777` by default, and SQLite creates databases as `0644`, both masked by the process umask. With a common `022` umask this leaves the hashed directory traversable and `userdata.db` globally readable even though it contains the service-device row and synced vault state. Please pre-create and verify the credential-specific leaf with owner-only permissions (for example `0700` on Unix) before spawning `dcli`; hashing the directory name does not protect its contents.

docs/src/content/docs/providers/dashlane.md
```diff
@@ -0,0 +1,205 @@
...
+
+**`secretspec set` fails.** Add or edit the item in a Dashlane app, run
+`dcli sync`, then read it here. Give the secret a writable provider as well if
+you need to set values from SecretSpec.
+
+**A new item is invisible until the CLI syncs.** `dcli` reads a local copy of
+the vault. SecretSpec never asks it to sync, but `dcli` syncs itself whenever
+its last sync is over an hour old, so a read is usually local and occasionally
+a network round-trip. Run `dcli sync` after adding an item rather than waiting
+for that, and `dcli configure disable-auto-sync true` to keep reads strictly
```
domenkozar: **[P2] Make the offline setting apply to credential-scoped state**

A normal `dcli configure disable-auto-sync true` updates the device row in the regular `dcli` database. For injected credentials, however, SecretSpec replaces `HOME` and `APPDATA` with the hidden hashed directory above, so the documented command edits a different database and SecretSpec reads still auto-sync. Please expose an offline provider option or otherwise apply the setting inside the same credential scope, or qualify these instructions so they do not promise strictly offline reads for injected credentials.

secretspec/src/provider/dashlane.rs
ojhermann: Fixed in 21286d4. `scoped_state_dir` now returns an error instead of `None`, and `run` propagates it, so an unresolvable cache directory aborts before the spawn rather than falling through to the inherited HOME.

secretspec/src/provider/dashlane.rs
ojhermann: Fixed in 21286d4. SecretSpec now creates the leaf itself with `DirBuilder::mode(0o700)` and tightens an existing one that has any group/other bits, so `dcli` writes into a directory no one else can enter. Checking the numbers first: against an empty HOME under a 022 umask, dcli 6.2628.1 leaves `dashlane-cli/` at 0755 and `userdata.db` at 0644, as you said. A unit test covers both the fresh and the already-0755 case.

docs/src/content/docs/providers/dashlane.md
ojhermann: Qualified in 21286d4 rather than made configurable. `dcli configure disable-auto-sync true` needs a login — against a fresh state directory it just prompts for an email — so the setting cannot be applied to a credential scope before that scope has registered and synced, and that scope necessarily starts empty. The provider page, the reference page and the struct docs now say the offline setting applies to a device you logged in yourself, and that injected credentials read from their own state and sync on dcli's hourly schedule.

PR #225, opened 2026-07-30
Workaround to make secretspec-derive work with as_path secrets

Redrield:
Right now, `secretspec-derive` discards the ValidatedSecrets after pulling out relevant values to construct its own Resolved to return to the user. This ignores the `temp_files` variable of ValidatedSecrets, holding `NamedTempFile` values whose Drop impls will delete the underlying temp files. Since that value isn't touched, and the ValidatedSecrets is dropped at the end of load(), temp files get deleted immediately after loading.

The workaround here isn't the best, but it was the easiest way I could think to do it since temp_files is pub(crate)- It stores ValidatedSecrets inside of Resolved, so that the temp_files only get dropped when the Resolved value does. 

Another way around this would be to make temp_files public so that derive instances can pull it out directly, rather than boxing and storing a copy of the secrets.


domenkozar:
Thanks for giving this one a shot! I've made #227 that fixes it a bit more structurally and comes with a test.

PR #236, opened 2026-07-31
fix(providers): reuse Vault-compatible logins per operation

dc-tec:
## Summary

- Reuse AppRole and JWT login tokens within each Vault/OpenBao provider operation.
- Share authentication across batched reads and multi-request writes or deletes.
- Re-authenticate when a token reaches its reported use or lease limit.
- Drop dynamic authentication state after each operation rather than caching or renewing tokens indefinitely.

## Rationale

Previously, AppRole and JWT authenticated for every underlying HTTP request, causing redundant login or OIDC requests during batch operations.

Operation-scoped reuse removes that overhead while remaining compatible with finite-use and short-lived tokens. Lease accounting includes time spent completing authentication, and invalid batch addresses are rejected before login so one-use AppRole SecretIDs are not unnecessarily consumed.

The shared HTTP execution also remains safe when called from multi-threaded or current-thread Tokio runtimes.

## Verification

- Full workspace test suite passes.
- Clippy and rustfmt pre-commit checks pass.
- Vault-only and OpenBao-only feature test suites pass.
- Regression tests cover finite-use tokens, lease expiry, delayed login responses, invalid batch addresses, and current-thread Tokio runtimes.
- Live OpenBao 2.6.1 AppRole smoke tests confirmed:
  - One login can serve a two-secret batch, with a fresh login for the next operation.
  - A token_num_uses=1 role re-authenticates as needed to complete a two-secret batch.

### DIFF
--- a/secretspec/src/provider/vault_common.rs
+++ b/secretspec/src/provider/vault_common.rs
@@ -10,13 +10,53 @@ use crate::{Result, SecretSpecError};
 use reqwest::header::{HeaderMap, HeaderValue};
 use secrecy::{ExposeSecret, SecretString};
 use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, VecDeque};
 use std::path::PathBuf;
 use std::sync::OnceLock;
+use std::time::{Duration, Instant};
+use tokio::sync::Mutex;
 use url::Url;
 
 pub(crate) const ROLE_ID: &str = "role_id";
 pub(crate) const SECRET_ID: &str = "secret_id";
 pub(crate) const TOKEN: &str = "token";
+
+/// Stable runtime for the shared Vault-compatible HTTP connection pools.
+///
+/// `get_many` invokes its synchronous fetch closure from several OS threads.
+/// Giving each closure a temporary runtime can strand a pooled reqwest
+/// connection when the runtime that owns its dispatch task is dropped. One
+/// process-wide runtime keeps those tasks alive across requests and providers.
+fn runtime() -> &'static tokio::runtime::Runtime {
+    static RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+
+    RUNTIME.get_or_init(|| {
+        tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("Failed to create Vault-compatible HTTP runtime")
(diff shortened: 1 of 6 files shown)


dc-tec:
Returned this PR to draft. Further review uncovered an issue with limited-use tokens: operation-scoped reuse can exhaust the token during batched reads or multi-request mutations, causing subsequent requests to fail with a 403. This needs to be resolved before the PR is ready for review.

secretspec/src/provider/vault_common.rs
```diff
@@ -369,12 +381,28 @@ impl KvProvider {
         self.http.get_or_init(reqwest::Client::new)
     }
 
+    /// Authenticates once for the logical operation that owns the returned
+    /// session. Dropping the session drops the token; a later operation logs in
+    /// again and therefore cannot unknowingly reuse an expired token.
+    fn session(&self) -> Result<KvSession<'_>> {
+        let token = super::block_on(self.resolve_token())?;
```
domenkozar: **[P1] Avoid reusing limited-use login tokens**

When an AppRole or JWT role sets `token_num_uses` below the number of HTTP requests in an operation, this session mints one token and reuses it throughout. The first requests consume the token, so later reads in `get_many` and the second request in `set_expiring` or `delete` receive 403 responses; these configurations previously worked because each underlying request performed a fresh login. Account for the returned token’s use limit and re-authenticate when necessary.

dc-tec: Yup, caught that (see my earlier comment) :)

Fix incoming.

dc-tec: Token reuse is now bounded by the server-reported auth.num_uses and auth.lease_duration, including time spent completing the login request. 

Exhausted or expired tokens trigger reauthentication, with regression coverage for batched reads and multi-request writes and deletes.

dc-tec:
Non-blocking follow-up: this work highlighted that `vault_common.rs` now combines configuration, authentication, token lifecycle, HTTP execution, KV operations, and their tests. 

Rather than mixing a structural refactor into this PR, would there be interest in a separate follow-up to move the shared implementation into `src/provider/vault_compatible/` with focused modules, while retaining `vault.rs` and `openbao.rs` as independent provider entry points?

PR #237, opened 2026-07-31
Add stored secret deletion and import cleanup

### LINKED ISSUE #229: Feature request: Secretspec clear
If I have run e.g. `secretspec import`, I probably do not want my secrets to continue to live around in the previous provider. I would like to be able to clear them all easily. Alternatively, I may have set a secret with `secretspec set` and later wish to remove the secret from the provider, without needing to dig into the details of said provider.

domenkozar:
## Summary

- add `secretspec delete <NAME>...` and a guarded `secretspec delete --all`
- expose idempotent deletion through `Secrets::delete`, following the primary write route and invalidating caches
- add `secretspec import <PROVIDER> --delete-source`, deleting source values only after destination verification
- preflight all migration routes, retain source values on conflicts, and reject unsafe same-store moves
- audit deletion operations and document the 0.18+ CLI/SDK behavior

## User impact

Users can remove values through SecretSpec without editing provider storage directly. Provider fallback copies are not implicitly destroyed, bulk deletion requires confirmation, and import cleanup behaves like a verified move rather than an unchecked copy-and-delete.

Closes #229.

## Validation

- `cargo test --all`
- `cargo clippy -p secretspec --no-default-features --features cli -- -D warnings`
- `npm run build` in `docs/`
- commit hooks: Clippy and rustfmt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Vault and OpenBao AppRole authentication now supports roles configured with
   `bind_secret_id=false` by omitting `secret_id` from the login request when no
   SecretID credential is configured.
+- `secretspec import --delete-source` now recognizes equivalent dotenv paths,
+  symlinks, and hard links as the same store, preventing source cleanup from
+  deleting the destination value.
 
 ### Added
 - Vault and OpenBao AppRole and JWT authentication can target non-default auth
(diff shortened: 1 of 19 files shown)


ap-1:
@domenkozar is there any possibility of adding a `secretspec sync` command that finds and deletes orphaned secrets no longer declared in `secretspec.toml`?

alternatively, a `--delete-undeclared` flag on `secretspec check` to be slightly more consistent with `secretspec import --delete-source`

would have to think about how this would work with `ref`, it's probably unsafe to delete secrets outside of secretspec's `{project}/{profile}/{key}` structure

domenkozar:
> @domenkozar is there any possibility of adding a `secretspec sync` command that finds and deletes orphaned secrets no longer declared in `secretspec.toml`?
> 
> alternatively, a `--delete-undeclared` flag on `secretspec check` to be slightly more consistent with `secretspec import --delete-source`
> 
> would have to think about how this would work with `ref`, it's probably unsafe to delete secrets outside of secretspec's `{project}/{profile}/{key}` structure

let's move this to another use by describing exactly what's needed

PR #250, opened 2026-08-03
ci(actions): pin external actions to commit SHAs

dc-tec:
## Summary

- Pin all 105 external GitHub Action invocations in hand-maintained workflows to full commit SHAs.
- Preserve the currently selected action versions, including the mixed checkout, download-artifact, and Cachix versions.
- Add weekly Dependabot maintenance for GitHub Actions with a seven-day cooldown.
- Leave the 16 action invocations in the cargo-dist-generated release workflow on their generated major-version tags and exclude that file from Dependabot.

## Rationale

GitHub recommends using full-length commit SHAs as the only immutable way to reference an external action:

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

Current `main` contains 121 external action invocations. One was already pinned; this PR pins 104 additional invocations. The remaining 16 mutable references are all in `.github/workflows/v-release.yml`, whose contents are generated and validated by cargo-dist 0.28.2.

## Dependabot rollout

A live inventory found eight action dependencies that are already eligible for major-version updates, covering 78 of the 105 hand-maintained invocations:

- `actions/checkout`: v4/v5 to v7
- `actions/download-artifact`: v4/v5 to v8
- `actions/upload-artifact`: v4 to v7
- `actions/setup-dotnet`: v4 to v6
- `actions/setup-go`: v5 to v7
- `actions/setup-node`: v4 to v7
- `actions/setup-python`: v5 to v7
- `cachix/cachix-action`: v16 to v17

Dependabot will likely propose approximately eight follow-up PRs. Updates are intentionally ungrouped initially so each action and major-version change can be reviewed and reverted independently. Updates to checkout, download-artifact, and Cachix are also expected to normalize their currently mixed versions.

The default limit permits five simultaneous version-update PRs. We can retain that default for the initial rollout and lower `open-pull-requests-limit` later if the first update wave is too noisy.

## Generated release workflow

`.github/workflows/v-release.yml` is generated by cargo-dist 0.28.2. Manual SHA pins cause `dist plan` to reject the workflow as out of date, so its 16 generated action references remain on mutable major-version tags in this PR.

The generated file is excluded from Dependabot to prevent automated edits that cargo-dist would reject. A follow-up cargo-dist upgrade can move these values into `[dist.github-action-commits]` and restore generator-native SHA pinning.

## Impact

There are no SecretSpec runtime or release-behavior changes. Hand-maintained workflows continue using the same action versions and now execute the exact commits behind those versions. Action upgrades can be reviewed independently through subsequent Dependabot PRs.

## Validation

- Confirmed 105 external action references use full 40-character SHAs.
- Confirmed the only 16 mutable external `uses:` references are in the cargo-dist-generated release workflow.
- Confirmed cargo-dist 0.28.2 completes `dist plan --output-format=json`.
- Validated the Dependabot cooldown and generated-workflow exclusion.
- `actionlint -shellcheck=`
- `devenv shell -- pre-commit run -a`
- `git diff --check`


dc-tec:
Reverted the edit in the cargo-dist-generated v-release.yaml. The 16 action invocations will remain on mutable major-version tags. Generator-native SHA pinning will follow separately alongside a cargo-dist upgrade.

domenkozar:
One test failure

dc-tec:
> One test failure

Should be good now

PR #265, opened 2026-08-04
Allow profiles to opt out of default inheritance

domenkozar:
## What changed

- add profile-level `inherit = false` under `[profiles.<name>.defaults]` (0.19+)
- make opted-out profiles standalone: they inherit neither default-profile declarations nor omitted fields
- preserve existing inheritance when the setting is omitted
- carry the policy through `extends`, runtime resolution, generated profile schemas, docs, and validation
- clarify how profile, scope, provider, and `extends` responsibilities differ

The SOPS write-preview work from the original branch is excluded because it shipped separately on `main`.

## Why

Some projects use `default` as their development base and want production to inherit most of it, while other profiles represent unrelated secret sets. The extending profile can now declare that it is standalone without annotating every default secret.

## Review feedback

This follows the requested shape:

```toml
[profiles.standalone.defaults]
inherit = false
```

## Validation

- `cargo test --workspace --all-features`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Profiles can opt out of inheriting `[profiles.default]` by setting
+  `inherit = false` in their profile defaults (0.19+), allowing standalone
+  secret sets alongside profiles that still share the default declarations.
 - Windows ARM64 CLI release artifacts (`aarch64-pc-windows-msvc`).
 - Provider aliases can define native `ref` templates and secrets can override
   coordinates per leaf alias with `refs`, so fallback providers and import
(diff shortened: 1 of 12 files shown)


alanscodelog:
I have only read your quick summary not the full changes, but regarding "add per-secret inherit = false for declarations in [profiles.default]" 

This does not really help for me personally for any of my secrets because I'm using default profile to mean "dev" and I have a prod profile extending "dev" which extends 99% of it's secrets. But it's the other profiles I don't want extending default, they have nothing to do with dev or prod secrets. Currently I am using a scope to further "narrow" them to only their secrets. 

I was thinking something more like this, where the *extending* profile says, hey no, don't inherit secrets from default, I'm a standalone profile or whatever you want to call it.
```
[profiles.normally-extending-profile.defaults]
inherit = false # do not inherit default's secrets.
```

PR #267, opened 2026-08-04
docs: front page feature showcase animation

vimjoyer:
image


domenkozar:
Let's keep the wording and "getting started" and  "concepts" buttons, but integrate the interactive example :)

PR #269, opened 2026-08-04
Support linking the SDKs against secretspec-ffi via pkg-config

sandydoo:
- Uses [cargo-c](https://github.com/lu-zero/cargo-c) to install the library, C headers, and compatible `.pc` files
- Ruby:  new `--enable-pkg-config` build flag (`gem install secretspec -- --enable-pkg-config`)
- Haskell: new `use-pkg-config` cabal flag  (`cabal build -f use-pkg-config`)
- Go: new `pkgconfig` build tag (`go build -tags static,pkgconfig`)
- Updated CI to statically link all three of the above SDKs via pkg-config


domenkozar:
Needs a rebase and CI fixes

PR #271, opened 2026-08-04
fix(providers): render the SOPS path template in one pass

ojhermann:
## Summary

- Render the SOPS path template in a single pass instead of two sequential `.replace()` calls.
- Add unit tests for `SopsPathPattern`, which had none.

## Rationale

`render` substituted `{project}` and then `{profile}` into its own output, so a value that looks like the other placeholder was substituted a second time. With a project literally named `{profile}`, `secrets/{project}/{profile}.yaml` resolved to `secrets/production/production.yaml` — a different file on disk than the one configured. Neither project nor profile names are validated, so both are user-controlled.

Walking the template once copies substituted values out rather than rescanning them. Unknown placeholders and unclosed braces are kept literal rather than panicking: `validate` rejects both, but `SopsPathPattern` also derives `Deserialize`, which does not run it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - The keyring provider no longer intermittently fails with a "No default store
   has been set" error when resolving multiple secrets concurrently.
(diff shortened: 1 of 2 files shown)


domenkozar:

PR #278, opened 2026-08-05
fix(lastpass): report the full item template from uri()

### LINKED ISSUE #272: lastpass: uri() drops the item template after the first path segment
`LastPassProvider::uri()` keeps only the first path segment of `folder_prefix`, so it doesn't round-trip what `TryFrom` parsed. The documented example shows it: `lastpass://Work/SecretSpec/{project}/{profile}/{key}` parses to that full template, and `uri()` reports `lastpass://Work` — which reparses as the template `Work`.

The effect looks confined to reporting: `uri()` feeds `provider_uri` in provenance, and cache identity comes from `cache_fingerprint` rather than the URI. So it's what the audit trail records, not what resolves.

I couldn't tell whether the truncation is deliberate — the `folder == "Shared"` branch looks like it means something specific, and there are no `uri()` tests for this provider. Happy to send a patch if it should emit the full prefix.

ojhermann:
## Summary

- `uri()` emits the whole item template, not just its first segment.
- Adds `entry_container_identity()`, matching keyring/gopass/pass now that the URI carries a template.
- Adds `uri()` tests for this provider, which had none.

## Rationale

`uri()` reported only the first segment of the item template, and that string is read back as a provider spec — `secretspec-derive` hands it to `set_provider`, and cached routes are fingerprinted with it. So a template read back as something else entirely: `lastpass://Shared/{project}/{profile}/{key}` as the default `secretspec/{project}/{profile}/{key}`, a different folder, and `lastpass://Work/TeamA/{key}` as the literal item `Work` — one item for every secret in the profile. Templates differing below their first segment also fingerprinted alike, so repointing a cached route kept serving the old template's values until they expired. (#272 called this reporting-only; that was wrong.)

Keyring had the same truncation until dd755d2, which fixed keyring and re-encoded lastpass's truncated folder without removing the truncation. This finishes that.

Two notes. The removed `folder == "Shared"` branch is what mapped `Shared/…` onto the default template; it dates to 1dad1c7, where keyring's `uri()` was an unconditional constant too, so I don't think it encodes LastPass semantics — happy to restore it if it does. And templates differing only by a literal `%20` versus a space still collide, since `URI_ENCODE_SET` doesn't escape `%`; that's shared with keyring and gopass and wants its own change.

Closes #272.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   resolves to the file you configured instead of a different one.
 - Invalid SOPS path templates are now rejected when loading serialized
   provider configurations instead of being accepted without validation.
+- The LastPass provider now reports its full item template rather than only the
+  first segment. A multi-segment template such as
+  `lastpass://Shared/{project}/{profile}/{key}` used to be reported as plain
+  `lastpass`, which reads back as the default `secretspec/{project}/{profile}/{key}`
+  template — a different folder — and `lastpass://Work/TeamA/{key}` read back as
+  the literal item `Work`, one item for every secret. Templates that differ
+  below their first segment are now distinguished, so repointing a cached route
+  at a new template invalidates its cached values instead of serving the old
+  ones until they expire. Single-segment templates are unaffected; cached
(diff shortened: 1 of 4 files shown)


domenkozar:

PR #287, opened 2026-08-05
ci(workflows): pin build container images

dc-tec:
## Summary

- pin the manylinux and musllinux build images to immutable digests
- pin the multi-architecture .NET 8 Alpine SDK image used for package testing
- retain readable image tags alongside each digest

This covers 11 references to five unique images across the .NET, FFI, Go, and Node workflows.

## Rationale

The workflows previously referenced mutable container tags. A registry publisher could therefore change the bytes used by pull-request and release builds without any corresponding repository change or review.

Zizmor reported eight high/high `unpinned-images` findings across the four matrix-container workflows. Pinning the resolved manifests makes image changes explicit and reviewable while retaining the original platform and compatibility tags.

No build behavior is intended to change beyond fixing the images to the versions the existing tags resolved to when this change was prepared.

## Maintenance tradeoff

This trades silent upstream image drift for manual refreshes.

Dependabot currently only extracts repository-style `uses:` dependencies from GitHub Actions workflows. It does not update job-container matrix values or image references embedded in `docker run` commands. Adding a Docker ecosystem entry would not cover these workflow files either.

There are five unique images behind the 11 references, so a refresh is mechanical but must currently be initiated manually. I have intentionally not added another dependency bot or prescribed an update schedule in this PR.

A reasonable initial policy could be to refresh these pins during release preparation, in response to relevant security notices, or on an agreed periodic cadence. If that proves burdensome, a dedicated updater or consistency check can follow.

[Dependabot support reference](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#github-actions)

## Validation

- resolved every tag-plus-digest reference directly from its registry
- verified the PyPA image architecture mappings:
  - x86_64 images resolve to `linux/amd64`
  - aarch64 images resolve to `linux/arm64`
- verified the pinned .NET manifest contains both amd64 and arm64 variants
- `devenv shell -- actionlint -shellcheck=`
- `devenv shell -- pre-commit run -a`
- reran zizmor against the four affected workflows; the eight high/high `unpinned-images` findings are no longer reported

Zizmor still emits four low-confidence `unpinned-images` findings because it cannot statically prove that `matrix.container` only resolves to the pinned matrix values. The concrete values are all digest-pinned.

### DIFF
--- a/.github/workflows/dotnet-package.yml
+++ b/.github/workflows/dotnet-package.yml
@@ -44,13 +44,13 @@ jobs:
           - rid: linux-x64
             target: x86_64-unknown-linux-gnu
             runner: ubuntu-latest
-            container: quay.io/pypa/manylinux_2_28_x86_64
+            container: quay.io/pypa/manylinux_2_28_x86_64:2026.08.05-1@sha256:e0b40ace8e818e96026eb47714b01998cbca022a6995797d0905474ce3e82ae8
             library: libsecretspec_ffi.so
             rustflags: -C strip=symbols
           - rid: linux-arm64
             target: aarch64-unknown-linux-gnu
             runner: ubuntu-24.04-arm
-            container: quay.io/pypa/manylinux_2_28_aarch64
+            container: quay.io/pypa/manylinux_2_28_aarch64:2026.08.05-1@sha256:f766b402889e40f439e7a3ee5788eef1aa3ef399d0110107d27419fa2ba9d905
             library: libsecretspec_ffi.so
             rustflags: -C strip=symbols
           - rid: osx-x64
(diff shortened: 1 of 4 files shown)


dc-tec:
This looks like a flaky failing test, could you retrigger the failing jobs?

domenkozar:
The two .NET Linux failures stop before the build in yum install -y libicu. Both pinned manylinux images refresh the enabled EPEL repository, whose mirrors returned stale metadata (404s and timeouts) on x64 and ARM64.

libicu is supplied by the AlmaLinux base repositories, so this step can avoid the unrelated EPEL dependency:

    run: yum --disablerepo=epel install -y libicu

I retried the failed jobs, but GitHub canceled the first retry while both jobs were still queued. This PR comes from a fork with maintainer edits disabled, so I cannot push the change directly.

PR #290, opened 2026-08-06
Add null provider for defaults and ephemeral generation

### LINKED ISSUE #276: Improve case where SecretSpec is used for non-secret environment variables
With the introduction of secretspec into my repos, I've stopped using gitignored .env files entirely, instead replacing them with git versioned secretspec.toml files. However, while the .env files did store secrets previously, they also stored some values that I would not consider as secret. For example, LOCAL_PORT = 8090. 

In secretspec, I use the same approach as in the https://secretspec.dev/reference/configuration/#complete-example for APP_NAME, but this requires a provider to be configured for the nonsecret - a value that will never be stored in the provider.

A null provider could help with this, along with potential other approaches.

I realise that there could be some overlap between this and #257, but I think this approach to eliminating .env files should certainly be encouraged. While _most_ configuration belongs in the configuration of the service itself, there are some variables which are ideal to end up in secretspec (as a replacement for .env) for one reason or another. An obvious example to me is SPRING_PROFILES_ACTIVE - we have different environments that we can run the service for, each with their own set of secrets. If I configure SPRING_PROFILES_ACTIVE in my secretspec toml appropriately, I then use `secretspec run -P staging -- mvn spring-boot:run`, while otherwise I would have to use `secretspec run -P staging -- mvn spring-boot:run -Dspring-boot.run.profiles=staging`, running the risk of using the wrong profile for the wrong secrets, more typing, and greater risk of typo.

domenkozar:
## Summary

- add a built-in `null://` provider that always reports values as missing and rejects ordinary writes
- let non-sensitive environment configuration resolve from committed manifest defaults without contacting secret storage
- add a `GeneratedValuePersistence` provider capability and make `null://` return generated values for one resolution without persisting them
- route `secretspec get` through the same generation executor as `run` and SDK resolution
- document the provider and all ephemeral-generation behavior as available in SecretSpec 0.19+

## Why

Projects replacing mixed-purpose `.env` files still need a clean home for shared, non-sensitive values such as `LOCAL_PORT` or `SPRING_PROFILES_ACTIVE`. They also need session secrets that can be freshly generated for each command without leaving a durable copy. Routing those declarations through `null` makes the manifest default or generator authoritative without polluting keyrings or shared secret backends.

## User impact

Users can set `providers = ["null"]` on declarations with defaults or enabled generation. Defaults resolve from the manifest, while generated values are minted once per materializing resolution and never written to provider storage. Stored fallback values still win before generation. Required declarations with neither a default nor generation remain missing, and explicit writes continue to report that the provider cannot store values.

Custom providers can override `generated_value_persistence`; storage providers retain the safe `Persist` default.

Closes #276.

## Validation

- `cargo test --all`
- `cargo test --package secretspec provider::null::tests`
- Clippy and rustfmt pre-commit hooks
- `git diff --check`

### DIFF
--- a/docs/src/content/docs/providers/null.md
+++ b/docs/src/content/docs/providers/null.md
@@ -1,15 +1,16 @@
 ---
 title: Null Provider
-description: Use committed manifest defaults without a storage backend
+description: Use committed defaults or ephemeral generation without a storage backend
 ---
 
 :::caution[Version compatibility]
 The `null` provider is added in SecretSpec 0.19.
 :::
 
-The null provider always reports that a value is missing. SecretSpec then
-uses the declaration's `default`. This is useful for non-sensitive environment
-configuration that belongs in `secretspec.toml`, not in a secret store.
+The null provider always reports that a value is missing. SecretSpec then uses
+the declaration's committed `default` or generates a fresh value when
+`generate` is enabled. This is useful for non-sensitive environment
+configuration and secrets that should exist for only one resolution.
 
 ## At a glance
 
@@ -17,8 +18,8 @@ configuration that belongs in `secretspec.toml`, not in a secret store.
 | --- | --- |
 | Provider | `null` (0.19+) |
 | URI | `null://` |
-| Access | Always returns missing; writes are rejected |
-| Best for | Team-shared, non-sensitive defaults |
+| Access | Always returns missing; ordinary writes are rejected |
+| Best for | Team-shared defaults and ephemeral generated values |
 | Storage | None |
 
 ## Quick start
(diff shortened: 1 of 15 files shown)


docs/src/content/docs/providers/null.md
```diff
@@ -0,0 +1,55 @@
...
+description: Use committed manifest defaults without a storage backend
+---
+
+:::caution[Version compatibility]
+The `null` provider is added in SecretSpec 0.19.
+:::
+
+The null provider always reports that a value is missing. SecretSpec then
+uses the declaration's `default`. This is useful for non-sensitive environment
+configuration that belongs in `secretspec.toml`, not in a secret store.
```
nmunnich: I would document here that this is also useful for secrets that are intended to always be generated.

moreau-nicolas:
Hi. I think that the provider should not only be able to let secretspec return manifest default values, but use the fallback providers set in the manifest too.

Given the following manifest :

```toml
[providers]
remote1 = "..."
remote2 = "..."

[profiles.default]
SOME_SECRET = { description = "Some secret" }

[profiles.development]
SOME_SECRET = { default = "a_shared_developer_secret_that_is_not_really_secret"  }

[profiles.staging]
SOME_SECRET = { providers = ["remote1"] }

[profiles.production]
SOME_SECRET = { providers = ["remote2"] }
```

When I run the following commands :
Then I would expect the following results :

 ```bash
$ secretspec get --provider null --profile development SOME_SECRET
# should get "a_shared_developer_secret_that_is_not_really_secret"

$ secretspec get --provider null --profile staging SOME_SECRET
# should get the value from the "remote1" provider

$ secretspec get --provider null --profile production SOME_SECRET
# should get the value from the "remote2" provider
```

If this behaviour is accepted, naming the provider `manifest` rather than `null` might be clearer.

File: docs/src/content/docs/providers/null.md (hunk 1 above)
Comment (domenkozar) [reply docs/src/content/docs/providers/null.md]: `null` rejects writes, including generated secrets. It does support defaults, because those never reach the secret provider.

We should add `memory` provider that doesn't persist.

**Comment by domenkozar** [comment]
Thanks for the suggestion. The profile-dependent behavior you describe can be
expressed with the current provider routing, but without `--provider null`:
`--provider` is an explicit override, so it intentionally replaces the provider
route declared in the manifest.

Instead, `null` can be the inherited route and each environment can override it
where needed:

```toml
[providers]
remote1 = "..."
remote2 = "..."

[profiles.default]
SOME_SECRET = { description = "Some secret", providers = ["null"] }

[profiles.development]
SOME_SECRET = { default = "a_shared_developer_secret_that_is_not_really_secret" }

[profiles.staging]
SOME_SECRET = { providers = ["remote1"] }

[profiles.production]
SOME_SECRET = { providers = ["remote2"] }
```

Then `secretspec get --profile <profile> SOME_SECRET` uses the manifest default
in development and the declared remote provider in staging or production.

I think `null` remains the more precise name for this implementation: the
provider itself only reports a missing value and does not inspect the manifest
or delegate to other providers. Calling it `manifest` would imply those broader
responsibilities.

PR #306, opened 2026-08-07
docs: rework getting started section

vimjoyer:
This reworks “getting started” to be more accessible by focusing on the key concepts one by one

### DIFF
--- a/docs/src/content/docs/common-commands.md
+++ b/docs/src/content/docs/common-commands.md
@@ -1,0 +1,108 @@
+---
+title: Common Commands
+description: The SecretSpec commands you will use most often
+---
+
+Once your project has a `secretspec.toml` file and you have selected a default
+provider, most day-to-day work uses a small set of commands.
+
+## Check required secrets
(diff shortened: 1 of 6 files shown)


docs/src/content/docs/common-commands.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Common Commands
```
domenkozar, line 2: Let's call this basic usage

docs/src/content/docs/migration.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Migration
+description: Bring existing secret declarations and values into SecretSpec
+---
+
+SecretSpec can discover declarations from an existing `.env` file or copy
+values from another provider. Secret values are never written to
+`secretspec.toml`.
+
+## Start a new project from `.env`
```
domenkozar, line 10: Note that `reflect()` is implemented for multiple providers, we should mention this!

PR #309, opened 2026-08-08
feat(provider): add sec provider via the sec CLI

Kaidstor:
-


domenkozar:
You'll have to gather some interest first before integrating with secretspec.

The README is not even in English.

I'll have to create some rules for adding providers, we can not add everyone's hobby project.

Kaidstor:
Fair enough — thanks for taking a look. I'll keep the provider maintained in my fork and come back if sec earns enough users to clear the bar. English README is on the way regardless. When you write down the rules for adding providers, happy to be a test case.

PR #317, opened 2026-08-10
perf(onepassword): batch field reads with op inject

mblarsen:
## Summary

- Batch multiple 1Password field references into one `op inject` process.
- Keep the existing whole-item batch path.
- Deduplicate identical references and copy each result to all request names.
- Preserve the current missing-secret behavior through an `op read` fallback.

## Motivation

SecretSpec previously started one `op read` process for each field reference. Large profiles therefore paid repeated process and unlock costs.

This implementation draws inspiration from [fnox](https://github.com/jdx/fnox) and [env-lease](https://github.com/mblarsen/env-lease), my own project. Both projects batch 1Password references through `op inject`.

## Implementation

The batch uses random framing markers around each reference. The parser uses these markers instead of JSON, because injected secret values can contain arbitrary JSON characters.

The parser preserves empty values, Unicode, spaces, backslashes, quotes, equals signs, and multiline values. It accepts the single trailing newline that `op inject` adds. It does not trim secret values.

If the `op inject` command fails, SecretSpec reads each unique reference separately. This fallback preserves missing-field handling and the existing provider errors.

A batch with one unique field reference still uses one `op read` call. Mixed batches keep the existing whole-item path and use `op inject` for field references.

The existing command runner still handles account selection, service-account tokens, session cleanup, stdin, and error translation.

## Scope

SecretSpec supports file-shaped output through `as_path = true`. It writes a resolved secret to an independent temporary file after provider resolution.

This output behavior is not full native 1Password file support. File attachments can use secret references, but standalone Document items require `op document get`.

The existing provider does not call `op document get`. Support for standalone Document items and native binary file transport remains outside this change.

## Performance

I tested a representative profile with 25 1Password field references. The test cleared the local cache before each measurement and did not log secret values.

| Version | Time | `op` processes | Exit |
| --- | ---: | ---: | ---: |
| Baseline | 96.294 s | 27 | 0 |
| This change | 11.890 s | 3 | 0 |

This run was 8.10 times faster and used 24 fewer `op` processes.

## Validation

- `cargo test --package secretspec onepassword`: 33 tests passed.
- `cargo fmt --all -- --check`: passed.
- Focused library clippy validation: passed.
- End-to-end 1Password QA: passed with three `op` processes.
- The workspace test run passed 1,179 tests.
- It also reported 21 SOPS test failures because the local system did not have `sops`.
- The workspace test excluded the PHP package because the local system did not have PHP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
+- 1Password field references now resolve in one batched CLI call, reducing
+  repeated unlocks and process startup when loading multiple secrets.
 - The Rust SDK's `ProviderAlias` now provides `leaf`, `credentials`, and
   `credentials_mut` helpers so callers can construct and inspect leaf or
   inline-cached aliases without depending on their storage representation.
--- a/secretspec/src/provider/onepassword.rs
+++ b/secretspec/src/provider/onepassword.rs
@@ -84,6 +84,74 @@ pub struct SecretReference {
     pub field: String,
 }
 
+/// Collision-resistant framing around each `op inject` expression.
+///
+/// Inject performs textual replacement, so formats such as JSON cannot safely
+/// carry arbitrary secret text without a format-aware escaping guarantee. These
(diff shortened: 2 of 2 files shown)


domenkozar:
Amazing work! :partying_face:

Issue #310, opened 2026-08-08
Create a SDK for JVM-languages


moreau-nicolas:
I think it would be great to support JVM-Languages.

I'm willing to contribute a PR.

domenkozar:
PR welcome!

moreau-nicolas:
I started working on it, targetting Java 11+ (2018 seems far enough : older versions are much less pleasant to work with).

Is the `setAsEnv()` method absolutely required ? The makers of Java had a pretty strong opinion about environment variables: they should be read-only. For this reason there is no standard way to set environment variables from within a running JVM program.

One alternative would be to use *system properties* which can be edited at runtime :

```java
SecretSpec.builder()
          .withPath("secretspec.toml")
          .withProvider("my-provider")
          .withReason("example")
          .load()
          .setAsSystemProperties(); // alternative to setAsEnv()
```

And later :

```java
var databaseUser = System.getProperty("DATABASE_USER"); // standard syntax.
```

domenkozar:
> I started working on it, targetting Java 11+ (2018 seems far enough : older versions are much less pleasant to work with).
> 
> Is the `setAsEnv()` method absolutely required ? The makers of Java had a pretty strong opinion about environment variables: they should be read-only. For this reason there is no standard way to set environment variables from within a running JVM program.
> 
> One alternative would be to use _system properties_ which can be edited at runtime :
> 
> SecretSpec.builder()
>           .withPath("secretspec.toml")
>           .withProvider("my-provider")
>           .withReason("example")
>           .load()
>           .setAsSystemProperties(); // alternative to setAsEnv()
> And later :
> 
> var databaseUser = System.getProperty("DATABASE_USER"); // standard syntax.

I'm fine if `setAsEnv` is not supported.

domenkozar:
https://github.com/cachix/secretspec/pull/318

moreau-nicolas:
@domenkozar I think the issue should not be closed yet : the artifact is not available on Maven Central.

domenkozar:
https://central.sonatype.com/artifact/org.cachix/secretspec-jvm/0.20.0

moreau-nicolas:
Thanks ! :)

PR #318, opened 2026-08-10
feat(sdk): secretspec-jvm

moreau-nicolas:
Add CI workflow.
Add doc.
Add examples.
Use gradle.
Handle MUSL libraries.
Update .gitignore.
Update scripts.
Update conformance.


moreau-nicolas:
Fixes #310.

GitHub Actions workflow still needs work.

domenkozar:
Sorry it took so long, CI is failing

gpampara:
I wonder if we need to do that xml prep step for the publishing vs just deferring to gradle to prepare the artifacts? It would be far less fragile and simpler to maintain going forward.

moreau-nicolas:
@gpampara The gradle plugins I found and tried expect to build locally before publishing, which would fail to include all the platform specific libraries. The XML format is pretty stable and should not pose any maintenance headache in the future.

@domenkozar Sorry for the failing build, it is my first GitHub Action workflow ever. There will probably be other mistakes. 😅

moreau-nicolas:
Build is passing now.

Next steps :
- [ ] Register an account with Maven Central to publish the artifacts (will require to confim ownership of cachix.org domain)
- [ ] Create GPG keys to sign the artifacts
- [ ] Add the newly created secrets to the GitHub Workflow
- [x] Update documentation to replace the 0.XX version placeholder with the target release

domenkozar:
> Build is passing now.
> 
> Next steps :
> 
> * [ ]  Register an account with Maven Central to publish the artifacts (will require to confim ownership of cachix.org domain)
> * [ ]  Create GPG keys to sign the artifacts
> * [ ]  Add the newly created secrets to the GitHub Workflow
> * [ ]  Update documentation to replace the 0.XX version placeholder with the target release

Is there another way to sign except GPG keys?

conformance/run.sh
```diff
@@ -111,6 +115,7 @@ else
   echo "==> SKIP Swift (the XCFramework SDK is macOS-only)"
   names+=("Swift"); statuses+=("SKIP")
 fi
+run "JVM"     jvm   run_jvm
```
domenkozar: **The JVM conformance suite never runs.**

`run()` guards on `command -v "$tool"`, and no binary named `jvm` exists on any machine, so this line always prints `==> SKIP JVM (jvm not found)`, records `SKIP`, and never calls `run_jvm`. The summary stays green while the JVM SDK is never checked against the shared fixtures.

Every other entry names a real executable (`run "C#" dotnet run_dotnet`). This should be `java`, or `gradle`.

.github/workflows/jvm-package.yml
```diff
@@ -0,0 +1,487 @@
...
+        run: |
+          SNAPSHOT_VERSION=$(grep 'secretspec.version' secretspec-jvm/gradle.properties | cut -d'=' -f2 | tr -d '[:space:]')
+          VERSION="${SNAPSHOT_VERSION%-SNAPSHOT}"
+          BUNDLE_DIR="bundle-dir/org/cachix/secretspec-jvm/${VERSION}"
+          mkdir -p "${BUNDLE_DIR}"
+          cp artifacts/secretspec-jvm-${SNAPSHOT_VERSION}.jar "${BUNDLE_DIR}/secretspec-jvm-${VERSION}.jar"
+          cp artifacts/secretspec-jvm-${SNAPSHOT_VERSION}-javadoc.jar "${BUNDLE_DIR}/secretspec-jvm-${VERSION}-javadoc.jar"
+          cp artifacts/secretspec-jvm-${SNAPSHOT_VERSION}-sources.jar "${BUNDLE_DIR}/secretspec-jvm-${VERSION}-sources.jar"
+          cat <<EOF > "${BUNDLE_DIR}/secretspec-jvm-${VERSION}.pom"
+          <project xmlns="http://maven.apache.org/POM/4.0.0"
```
domenkozar: **The published POM declares no dependencies, but the jar is not shaded.**

`library/build.gradle.kts` has `implementation("net.java.dev.jna:jna:5.19.1")` and `implementation("com.fasterxml.jackson.core:jackson-databind:2.21.5")`, and there is no shadow/relocation step. This hand written POM lists groupId, artifactId, version, licenses, and scm only.

A consumer who adds `org.cachix:secretspec-jvm` and calls `SecretSpec.builder().load()` gets `NoClassDefFoundError: com/sun/jna/Library` (or `com/fasterxml/jackson/databind/ObjectMapper`).

The consumer smoke test cannot catch this: `consumer/build.gradle.kts` adds `testImplementation("net.java.dev.jna:jna")` itself, and only calls `SecretSpec.abiVersion()`, which never touches Jackson.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/JsonContracts.java
```diff
@@ -0,0 +1,412 @@
...
+            && Objects.equals(provider, other.provider)
+            && Objects.equals(profile, other.profile)
+            && Objects.equals(scope, other.scope)
+            && Objects.equals(secrets, other.secrets)
+            && Objects.equals(missingRequired, other.missingRequired)
+            && Objects.equals(missingOptional, other.missingOptional);
+    }
+}
+
+final class ReportResponseContract {
```
domenkozar, line 351: **`report()` throws on exactly the case it exists to describe.**

`ReportResponseContract` has no `constraint_violations` field, and `SecretSpecJsonContext.MAPPER` (JsonContracts.java:25) leaves Jackson's default `FAIL_ON_UNKNOWN_PROPERTIES=true`.

`secretspec/src/report.rs:88` serializes `constraint_violations` whenever the vector is non empty (0.17+). So a manifest with a failing cross secret presence constraint returns `{"ok":true,"response":{...,"constraint_violations":[...]}}`, `MAPPER.readValue` raises `UnrecognizedPropertyException`, and `SecretSpecBuilder.call` turns it into `SecretSpecException("parse", ...)`.

Two fixes are worth making together: add the field, and disable `FAIL_ON_UNKNOWN_PROPERTIES` on the shared mapper. The C#, Go, and Python SDKs all tolerate unknown fields, so as written this SDK also breaks on any future response field.

devenv.nix
```diff
@@ -112,6 +119,7 @@
       CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER = muslcc;
       MUSL_CC = muslcc;
       MUSL_STATIC_LDFLAGS = "-L${pkgs.pkgsStatic.libunwind}/lib";
+      SECRETSPEC_JVM_TARGET_JDK = "${pkgs.jdk11}";
```
domenkozar, line 124: **macOS contributors cannot build the JVM SDK.**

`env = lib.optionalAttrs pkgs.stdenv.isLinux (...)` at line 113, whose comment explains it is deliberately Linux only, so `SECRETSPEC_JVM_TARGET_JDK` is unset on macOS.

`gradle.properties` sets `auto-download=false` and `fromEnv=SECRETSPEC_JVM_TARGET_JDK,JAVA_HOME_11_X64,JAVA_HOME_11_ARM64`. None of those exist in a macOS `devenv shell`, so `./gradlew build` fails with "No matching toolchains found for Java 11".

The JDK is not part of the musl cross toolchain concern, so it can move out of the `optionalAttrs` block.

domenkozar (commented):
Review of the JVM SDK by [Claude Code](https://claude.com/claude-code) at the request of @domenkozar. Findings are inline.

The highest signal ones, roughly in order:

1. `conformance/run.sh` registers the JVM suite under a tool name (`jvm`) that does not exist, so it always SKIPs.
2. The Maven Central POM declares no dependencies, but the jar is not shaded, so the published artifact is unusable.
3. `report()` throws on any manifest with a failing constraint, because `constraint_violations` (0.17+) is missing from the contract and unknown fields are fatal.
4. `SECRETSPEC_JVM_TARGET_JDK` lands in the Linux only `env` block, so macOS contributors cannot build.
5. The JNA request string is encoded with the platform charset while the response is decoded as UTF-8.

Plus a set of packaging and release visibility items (version placeholders, missing CHANGELOG entry, unpinned Gradle wrapper, `curl` without `--fail`).

I verified each of these against the PR branch and, where relevant, against the CI logs for the current run. Two candidate findings did not survive that check and are not included: the `archive: false` uploads are fine (the file name becomes the artifact name, and I confirmed the flat download layout in the consumer job log), and `:library:javadoc` passes today.

Lower severity items not worth their own inline comment:

* `SecretSpecJsonContext.SECRET_FIELDS` is dead code.
* `Resolved` builds a second `ObjectMapper` instead of reusing `SecretSpecJsonContext.MAPPER`, and `fields()` iterates `keySet()` + `get()` + `Optional` instead of `entrySet()`.
* `Resolved`'s constructor uses `Map.copyOf` / `List.copyOf`, which NPE rather than raising `SecretSpecException` if the envelope ever omits `secrets` / `missing_optional`.
* `shouldUseMusl()` forks `ldd` at class init, never `waitFor`s, and leaks the second stream on the early `return true`.
* `NativeTest.abi_version_should_be_defined` duplicates `SecretSpecTest.testAbiVersion`; `testSetAsSystemProperties` never clears `DATABASE_URL` when there was no previous value; a stray `System.out.println(noValues)` remains in `testConformance`.
* The builder lacks `withCaller` / `CallerContext` (0.20+) that the .NET builder has.
* The Alpine consumer step nests single quotes inside a single quoted `sh -c`; the .NET workflow escapes double quotes there instead.
* `devenv.nix` adds both `pkgs.jdk11` and `pkgs.jdk21` to `packages` on top of `languages.java.jdk.package`. The CI log shows the result: `pkgs.buildEnv warning: colliding subpath (ignored): .../openjdk-11.../include/jvmti.h and .../openjdk-21.../include/jvmti.h`.
* `pre-release.yml` gained a trailing whitespace line, and its `jvm` job `paths:` trigger omits `secretspec-jvm/gradle.properties`.

<sub>Posted via `gh api` because the GitHub CLI cannot create inline review comments natively; see https://github.com/cli/cli/issues/13904.</sub>

secretspec-jvm/library/src/main/java/org/cachix/secretspec/Native.java
```diff
@@ -0,0 +1,198 @@
...
+
+
+final class Native {
+
+    private static final String LIBRARY_NAME = "secretspec_ffi";
+    private static final String MUSL_LIBRARY_NAME = "secretspec_musl_ffi";
+    private static final SecretSpecFFI BINDINGS;
+
+    interface SecretSpecFFI extends Library {
+        Pointer secretspec_resolve(String requestJson);
```
domenkozar: **The request is written with the platform charset while the response is read as UTF-8.**

JNA marshals a `String` parameter with `Native.getDefaultStringEncoding()`, which is `jna.encoding` if set, otherwise `Charset.defaultCharset()`. This SDK targets JDK 11, where the default charset still follows the platform locale: windows-1252 on Windows, US-ASCII under a POSIX/C locale container.

So `withPath("/srv/приложение/secretspec.toml")` or `withReason("déploiement")` sends bytes that are not valid UTF-8, and `resolve_inner` answers `{"ok":false,"error":{"kind":"invalid_input","message":"request_json was not valid UTF-8"}}`. In the lossy cases it can silently resolve a mangled path instead.

Line 60 already pins the read direction with `getString(0, "UTF-8")`. Pin the write direction too, by passing `Library.OPTION_STRING_ENCODING = "UTF-8"` to `com.sun.jna.Native.load`.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/Native.java
```diff
@@ -0,0 +1,198 @@
...
+                 * JNA does not distinguish between musl and glibc.
+                 * To ship two binaries targeting the same linux-arch in the Jar we must rename one of them.
+                 * We rename the musl library.
+                 */
+                BINDINGS = com.sun.jna.Native.load(MUSL_LIBRARY_NAME, SecretSpecFFI.class);
+            } else {
+                BINDINGS = com.sun.jna.Native.load(LIBRARY_NAME, SecretSpecFFI.class);
+            }
+        } catch (UnsatisfiedLinkError error) {
+            throw new SecretSpecException("load", error.getMessage(), error);
```
domenkozar: **The documented `kind = "load"` error is never observable.**

Throwing from a static initializer makes the JVM wrap it in `ExceptionInInitializerError`, an `Error` that `catch (SecretSpecException)` does not match, and every later touch of this class throws `NoClassDefFoundError: Could not initialize class org.cachix.secretspec.Native` with the original cause gone.

So on a platform whose asset is absent, or with `SECRETSPEC_FFI_LIB` pointing at a bad file, a caller of `SecretSpec.builder().load()` sees `NoClassDefFoundError`, not the contract that both the README and `docs/src/content/docs/sdk/jvm.mdx` promise: "Other failures throw `SecretSpecException`, whose `kind` property is a stable error category."

Loading lazily behind a holder class or a synchronized accessor keeps the exception on the calling path.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/Native.java
```diff
@@ -0,0 +1,198 @@
...
+        String fileName;
+        if (os.contains("win")) {
+            fileName = "secretspec_ffi.dll";
+        } else if (os.contains("mac")) {
+            fileName = "libsecretspec_ffi.dylib";
+        } else {
+            fileName = "libsecretspec_ffi.so";
+        }
+
+        String[] starts = new String[]{
```
domenkozar: **In production this walks every ancestor of the working directory and prefers whatever `target/` build it finds over the packaged Jar asset.**

The loop starts from `user.dir` and the code source directory and ascends to the filesystem root, returning the first `target/release|debug/libsecretspec_ffi.so` it hits. Nothing gates it on a development mode.

A packaged application launched from any directory underneath an unrelated Rust checkout, or with a writable `/target` planted by another user, silently `dlopen`s that library instead of the vetted asset in the Jar, and executes it on the secrets path.

The doc comment calls this "for local SDK development", so the intent is clear; it just needs a gate (an explicit env var, or restricting it to the case where the class is not loaded from a Jar). It also weakens the CI consumer job, which can pass against a stray dev build rather than the packaged asset.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/Resolved.java
```diff
@@ -0,0 +1,164 @@
...
+
+        disposed = true;
+        UncheckedIOException firstError = null;
+        for (var secret : secrets.values()) {
+            if (!secret.isAsPath() || secret.getPath() == null)
+                continue;
+
+            try {
+                var path = Paths.get(secret.getPath());
+                Files.delete(path);
```
domenkozar: **`close()` throws when the temp file is already gone, which diverges from the cross SDK contract.**

`secretspec-go/secretspec.go:156` states "a file already gone is not an error" and skips `os.IsNotExist`, and .NET's `File.Delete`, which this class is ported from, is a no op for a missing path.

`Files.delete` instead throws `NoSuchFileException`, which line 128 wraps in `UncheckedIOException` and rethrows. So an `as_path` temp file reaped by a tmp cleaner, removed by the application after use, or dropped by a container restart makes a `try (var resolved = ...)` block that already succeeded fail at scope exit, and suppress any in flight exception.

`Files.deleteIfExists` restores the documented behaviour.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/ResolvedSecret.java
```diff
@@ -0,0 +1,92 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+
+/**
+ * One resolved secret and its provenance.
+ */
+public final class ResolvedSecret {
+
+    /**
+     * The inline value, or null for an <c>as_path</c> secret.
```
domenkozar: Minor: `<c>as_path</c>` is C# XML doc markup left over from the .NET port. Javadoc has no `<c>` element, so this renders without the code formatting it was meant to carry. Java wants `{@code as_path}`.

(I checked the CI logs: `:library:javadoc` currently passes, so this is cosmetic, not a build break.)

secretspec-jvm/examples/typed_access/TypedAccess.java
```diff
@@ -0,0 +1,12 @@
+import org.cachix.secretspec.SecretSpec;
+import io.quicktype.Converter;
+
+public class TypedAccess {
+
+    public static void main() {
+        try(var resolved = SecretSpec.builder().load()) {
+            AppSecrets typed = Converter.fromJsonString(resolved.fieldsJson());
```
domenkozar: **The examples do not compile, and nothing compiles them.**

This file is in the default package and imports only `io.quicktype.Converter`, so `AppSecrets` (which is `package io.quicktype`) does not resolve. And `Converter.fromJsonString` is declared `throws IOException` while `main()` declares no `throws`. Two compile errors.

All five examples also declare `public static void main()` with no `String[]` parameter, which is not a valid entry point on the JDK 11 target the package advertises. And `Scopes.java` has an empty `try` body, so it demonstrates nothing, although `docs/src/content/docs/sdk/jvm.mdx` presents it as the example for `Resolved.getScope()`.

This matters because every one of these files is embedded verbatim into the published SDK docs via `?raw` imports.

Nothing catches it today: `settings.gradle.kts` includes only `library` (plus `consumer`, conditionally), and `scripts/ci-sdks.sh:run_jvm` runs `./gradlew build` only. Compare `run_dotnet`, which does `find examples -name '*.csproj' -exec dotnet build {} \;`.

.github/workflows/jvm-package.yml
```diff
@@ -0,0 +1,487 @@
...
+          echo "$MAVEN_CENTRAL_SIGNING_KEY" | gpg --batch --import
+          # Ultimately trust the imported key
+          KEY_ID=$(gpg --with-colons --list-secret-keys | grep '^sec' | cut -d':' -f5)
+          echo "${KEY_ID}:6:" | gpg --import-ownertrust
+      - name: Prepare Maven Bundle Structure
+        shell: bash
+        env:
+          GPG_PASSPHRASE: ${{ secrets.MAVEN_CENTRAL_SIGNING_PASSPHRASE }}
+        run: |
+          SNAPSHOT_VERSION=$(grep 'secretspec.version' secretspec-jvm/gradle.properties | cut -d'=' -f2 | tr -d '[:space:]')
```
domenkozar, line 429: **The publish job reads a version the rest of the workflow has already rewritten.**

`native` (line 90), `musl` (line 185), and `package` (line 253) all run `scripts/sync-sdk-versions.sh`, which rewrites `secretspec.version` to the Cargo workspace version. This job skips that step and greps the checked in value, currently `0.19.1-SNAPSHOT`.

On a `v0.20.0` tag whose `gradle.properties` was not hand bumped, `package` produces `secretspec-jvm-0.20.0-SNAPSHOT.jar` while this computes `SNAPSHOT_VERSION=0.19.1-SNAPSHOT`, so the `cp` on line 431 fails, or worse publishes the wrong coordinates if that name happened to exist.

`docs/src/content/docs/development/sdks.md:34` says these versions "are not hand-edited: release workflows run scripts/sync-sdk-versions.sh", so adding the sync step here matches the documented contract.

.github/workflows/jvm-package.yml
```diff
@@ -0,0 +1,487 @@
...
+          zip -r ../maven-central-bundle.zip .
+          cd ..
+
+      - name: Upload to Maven Central via API
+        env:
+          MAVEN_CENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
+          MAVEN_CENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_TOKEN_PASSWORD }}
+        run: |
+          ENCODED_AUTH=$(echo -n "${MAVEN_CENTRAL_USERNAME}:${MAVEN_CENTRAL_PASSWORD}" | base64 -w 0)
+          curl --request POST \
```
domenkozar: **A rejected upload leaves this step green.**

`curl` exits 0 for any HTTP status unless you pass `--fail` (or `--fail-with-body`). An expired or rotated `MAVEN_CENTRAL_TOKEN_*` gives 401, and a bundle Sonatype rejects gives 400; in both cases the step succeeds, the release workflow reports success, and nothing is published. The comparable `dotnet nuget push` in `dotnet-package.yml` fails loudly.

Two related notes on this job:
* The `permissions:` block on line 400 comments `id-token: write # no long-lived key`, but this job authenticates with long lived `MAVEN_CENTRAL_TOKEN_USERNAME` / `_PASSWORD` and a GPG signing key. The comment describes a trusted publishing setup that is not in use here.
* Unlike `dotnet-package.yml:492`, there is no `environment:` gate on the job holding those secrets.

secretspec-jvm/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
```
domenkozar, line 3: **The Gradle wrapper is the one unpinned executable in the release pipeline.**

`validateDistributionUrl=true` only validates the URL's host, not the archive's contents, so there is no integrity check on `gradle-8.14.4-bin.zip`. And the 43 KB `gradle/wrapper/gradle-wrapper.jar` committed alongside this file is executed by every job and is unreviewable in a diff.

Everything else in this repo is pinned: every GitHub Action is referenced by commit SHA, and the workflow installs a "verified rustup". A tampered wrapper jar or a compromised distribution here runs inside the publish job that holds `MAVEN_CENTRAL_SIGNING_KEY` and the publishing token.

Two cheap fixes: add `distributionSha256Sum` here, and add a `gradle/actions/wrapper-validation` step to the workflow.

docs/src/content/docs/sdk/jvm.mdx
```diff
@@ -0,0 +1,82 @@
...
+---
+
+import { Code } from 'astro:components';
+import quickStartExample from '../../../../../secretspec-jvm/examples/quick_start/QuickStart.java?raw';
+import scopesExample from '../../../../../secretspec-jvm/examples/scopes/Scopes.java?raw';
+import reportExample from '../../../../../secretspec-jvm/examples/report/Report.java?raw';
+import typedAccessExample from '../../../../../secretspec-jvm/examples/typed_access/TypedAccess.java?raw';
+import asPathExample from '../../../../../secretspec-jvm/examples/as_path/AsPath.java?raw';
+
+> **Version compatibility:** Available since SecretSpec 0.XX.
```
domenkozar: **Release visibility and listing updates are incomplete.**

CLAUDE.md: "Any unreleased provider, CLI command, configuration field, or syntax must be labeled with its target version at each point of use." The docs site publishes from `main`, so readers currently see the literal placeholder:

* here and at line 19 (`## Install (0.XX+)`)
* `docs/astro.config.ts:433` sidebar badge
* `docs/src/content/docs/sdk/overview.md:38,66,139`
* `conformance/README.md:67`
* `secretspec-jvm/README.md:3`

The next release is 0.20: Cargo.toml is at 0.19.1 and the Unreleased section already labels the Azure provider `0.20+`.

CLAUDE.md also asks for a CHANGELOG entry under Unreleased. There is none, though both previous SDK additions have one (Swift and C#).

Stale listings elsewhere:
* `docs/src/pages/index.astro:20` derives `sdkCount` from the `sdk/` collection, so the count on lines 456 and 1888 silently increments to include JVM while the prose list on line 1888 and the consumer tile array on line 116 still stop at Swift.
* `docs/src/content/docs/development/sdks.md` is missing JVM rows in the packaging workflow table (line 50) and the platform support table (line 67). Its own "Adding a new SDK" checklist, step 5, requires them.
* `secretspec/README.md:281` "Language SDKs" has no JVM bullet.
* `sdk/overview.md` still says only "the C# package additionally ships musl Linux assets", although the JVM package now ships them too.

Finally, `## Install (0.XX+)` on line 19 has no Maven or Gradle dependency snippet, unlike `csharp.mdx:25`, so a reader cannot actually install the package from this page.

conformance/run.sh
moreau-nicolas: I'll use "bash": the gradle wraper downloads gradle.

.github/workflows/jvm-package.yml
moreau-nicolas: Good catch!

moreau-nicolas:
> Is there another way to sign except GPG keys?

Not that I know of.
Signing the artifacts is a requirement.
See [the Sonatype documentation](https://central.sonatype.org/publish/requirements/gpg/).
Is this a problem?

domenkozar (commented):
Supplemental JVM SDK review: two additional findings are inline. The remaining findings from this review are already represented by existing threads, so I did not duplicate them.

secretspec-jvm/library/src/main/java/org/cachix/secretspec/JsonContracts.java
```diff
@@ -0,0 +1,412 @@
...
+    private String scope;
+
+    @JsonProperty("reason")
+    private String reason;
+
+    @JsonProperty("no_values")
+    private boolean noValues;
+
+    @JsonProperty("mode")
+    private String mode;
```
domenkozar: **The JVM request builder cannot send the structured caller context supported by the 0.20 FFI contract.**

The native request accepts a `caller` object for audit metadata, and the other SDK builders expose its name, operation, and resource fields. `ResolveRequest` and `SecretSpecBuilder` only expose the legacy scalar fields, so JVM integrations cannot identify themselves in audit records.

Add a caller request contract plus a fluent builder method that serializes it as `caller`.

.github/workflows/jvm-package.yml
```diff
@@ -0,0 +1,487 @@
...
+        if: matrix.container
+        shell: bash
+        run: |
+          bash scripts/install-rustup.sh
+          echo "$HOME/.cargo/bin" >> "$GITHUB_PATH"
+
+      - name: Install Rust
+        run: rustup toolchain install
+
+      - uses: actions/setup-java@v5
```
domenkozar: **All `actions/setup-java` steps use the mutable `v5` tag in a release workflow.**

These steps install the JDK used to construct the JAR that is later signed and published. A moved or compromised tag can therefore change release artifacts without a reviewed repository change. The checkout and artifact actions in this workflow are already pinned; pin every `setup-java` occurrence to a reviewed commit SHA as well.

PR #324, opened 2026-08-10
docs: validate internal links

vimjoyer:
This should protect us against dead internal links (we had one)


domenkozar:
Let's use some existing linkchecker like https://github.com/lycheeverse/lychee

PR #325, opened 2026-08-10
docs: standardize provider guides

vimjoyer:
This aims to make bw and sops follow the general style of the other provider guides. Also fixes an issue where bw page thought we are still on 0.17

### DIFF
--- a/docs/src/content/docs/providers/bw.md
+++ b/docs/src/content/docs/providers/bw.md
@@ -3,14 +3,47 @@ title: Bitwarden Password Manager Provider
 description: Bitwarden Password Manager secrets management integration
 ---
 
-The `bw` provider integrates with Bitwarden Password Manager for secret management with vault-wide access to all item types.
+The `bw` provider reads and writes secrets in Bitwarden Password Manager by
+using the official `bw` CLI.
 
 :::note[Version compatibility]
-The Bitwarden Password Manager provider is an upcoming SecretSpec 0.18 feature and is not
-available in SecretSpec 0.17.
+The Bitwarden Password Manager provider was added in SecretSpec 0.18.
 :::
 
-## Prerequisites
+## At a glance
+
+| | |
+|---|---|
+| Provider | `bw` |
+| URI | `bw://[COLLECTION\|ORGANIZATION@COLLECTION][?options]` |
+| Access | Read and write |
+| Best for | Existing Bitwarden Password Manager vaults and items |
+| Authentication | An unlocked `bw` CLI session through `BW_SESSION` |
+| Build feature | `bw` |
+
+## Quick start
+
+Sign in, unlock the vault, and export the session returned by `bw unlock`:
+
+```bash
+$ bw login
+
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/providers/sops.mdx
```diff
@@ -1,58 +1,148 @@
...
+  your package
 
-### URI Format
+For a new file, SOPS needs either encryption options in the provider URI or a
+matching creation rule in `.sops.yaml`. For example:
 
+```yaml title=".sops.yaml"
+creation_rules:
+  - path_regex: secrets\.enc\.yaml$
+    age: "age1jpa8rf5qmrg6pw444fcgpkaxg8x4neueszrexzagdjpunjlgeyzq304w34"
```
domenkozar: **[P1] Replace the published test recipient in the setup rule**

When a user copies this creation rule, SOPS encrypts to `age1jpa8rf5...`, whose matching private key is committed at `secretspec/src/provider/sops/test_fixtures/key.txt`. Any secret encrypted with this rule is therefore decryptable by anyone with the repository; generate a user-owned key and recipient in the setup instead of publishing the fixture recipient.

docs/src/content/docs/providers/bw.md
```diff
@@ -152,11 +204,34 @@ context does not change Bitwarden item names. To migrate the discovered values
 after reviewing the declarations, run the `secretspec import` command printed
 by `init`.
 
-### Item Type Configuration
+### Environment defaults
+
+Environment variables can supply defaults when they are not present in the
+provider URI:
```
domenkozar: **[P1] Document that Bitwarden environment values override the URI**

When any of these `BITWARDEN_*` variables is present alongside an explicit `bw://...` URI, the provider reads the environment first rather than only filling missing URI values. As written, a user can expect `bw://prod@collection` to win while a stale environment value silently redirects a read or write to another collection, item type, or field, so this section needs to state the actual precedence.

docs/src/content/docs/providers/bw.md
```diff
@@ -3,26 +3,107 @@ title: Bitwarden Password Manager Provider
...
+
+The provider uses the active `bw` CLI session. Export the session key after
+unlocking the vault:
+
+```bash
+$ export BW_SESSION="your-session-key"
+```
+
+SecretSpec distinguishes between a CLI that is signed out and a vault that is
+still locked, and reports the appropriate `bw login` or `bw unlock` command.
```
domenkozar: **[P2] Remove the auth-state-specific error promise**

With current `bw` versions, both signed-out and locked states are returned as successful JSON from `bw status`; `is_authenticated` collapses them to `false`, and the read, write, and discovery paths all emit the same combined `bw login`/`bw unlock` message. Users therefore do not receive the appropriate single command promised here, so either document the combined guidance or preserve the status through the preflight.

domenkozar (commented):
Three documentation issues are noted inline.

PR #328, opened 2026-08-12
Age provider delete

GrafBlutwurst:
Adds delete capability to the age provider so it can serve as a caching capable provider.


domenkozar:
Thank you!

PR #330, opened 2026-08-12
cli: add shell completions

vimjoyer:
image

Adds secretspec completions for Bash, Elvish, Fish, Nushell, PowerShell, and Zsh.

Completions are generated directly from the Clap command definition and include nested commands, options, possible values, and descriptions where supported. Includes tests and setup documentation.


vimjoyer:
image
completes secrets from the secretspec.toml and any other stuff

PR #341, opened 2026-08-14
fix: remove stale proc-macro-error2 crates-io git patch

ifiokjr:
## Why

The `.cargo/config.toml` shipped in the repo root declares a `[patch.crates-io]` section pointing at a git fork of `proc-macro-error2` (`github.com/GnomedDev/proc-macro-error-2`). Nothing in the dependency graph uses these crates — the `Cargo.lock` records them as `[[patch.unused]]` — but cargo still tries to fetch the git source during resolution.

That breaks **offline vendored builds**: nixpkgs' `buildRustPackage` builds with `--offline` and fails with:

```
error: failed to load source for dependency `proc-macro-error-attr2`
Caused by:
  can't checkout from 'https://github.com/GnomedDev/proc-macro-error-2.git': you are in the offline mode (--offline)
```

This was hit when packaging monosecret 0.2.0 in the ifiokjr/nixpkgs flake.

## What changed

- **Delete `.cargo/config.toml`** — the file contained only the stale patch section.
- **Prune `[[patch.unused]]` entries from `Cargo.lock`** — the two `proc-macro-error2` / `proc-macro-error-attr2` git entries.

## Verification

- `cargo metadata --locked` — resolves cleanly, no git sources in the graph
- `cargo tree -p monosecret -i proc-macro-error2` — not present (only the unrelated registry `proc-macro-error3` used by trybuild remains)
- `cargo check -p monosecret` — passes
- Changeset included (`.changeset/1786702621-monosecret.md`, patch bump)

The upstream comment in the removed config said "Remove after proc-macro-error2 publishes the public proc_macro re-export fix" — that fix has landed, and nothing in the tree depends on those crates anymore, so the patch is safe to drop.


ifiokjr:
Wrong base repo - reopening against ifiokjr/monosecret

PR #346, opened 2026-08-14
fix(onepassword): keep batch reads with absent items

domenkozar:
## Summary

- inventory referenced 1Password vaults after a failed `op inject`
- omit references whose item ID or title is absent, then retry all remaining references in one batch
- preserve known authentication and installation errors immediately instead of fanning out into individual reads
- retain the bounded individual-read fallback for missing fields on items that exist

## Root cause

`op inject` fails the entire template when any referenced item or field is missing. SecretSpec previously discarded that batch error and retried every unique reference with `op read`, so a few unprovisioned optional items permanently disabled batching for the whole profile.

## Impact

For absent optional items in one vault, the recovery path is now one failed inject, one structured item inventory, and one retry inject for all present references. This avoids spawning one CLI process per present reference and does not depend on matching 1Password's human-readable missing-item error text.

## Validation

- `cargo test --package secretspec provider::onepassword::tests --no-fail-fast` (32 passed)
- `cargo test --all`
- pre-commit `clippy` and `rustfmt` hooks

A live-vault benchmark was not run because the local 1Password desktop session was unavailable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   configuration metadata only; it never queries providers or reads secret
   values.
 
+### Fixed
+
+- 1Password batches no longer fall back to one CLI process per reference when
+  optional items are absent. SecretSpec inventories the referenced vaults,
+  omits absent items, and retries the remaining references as one batch. An
+  authentication failure now stops before fallback instead of fanning out into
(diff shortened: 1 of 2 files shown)


domenkozar:
Superseded by #347, which carries the same missing-item batch recovery with live benchmark coverage and broader edge-case tests. I added the remaining auth/install fail-fast hardening there in c785887, and its full CI matrix is green.

PR #350, opened 2026-08-15
Fix provider path boundary joining

### LINKED ISSUE #344: awssm: prefix with a trailing slash builds a double slash and silently finds nothing
## Summary

`awssm://<region>?prefix=myteam/` with a trailing slash builds a secret name containing a double
slash, and the lookup then fails with a plain "not found". Nothing points at the prefix as the
cause.

## Detail

`format_secret_name` in `secretspec/src/provider/awssm.rs` inserts the separator itself:

```rust
let secret_name = match prefix {
    Some(p) => format!("{}/secretspec/{}/{}/{}", p, project, profile, key),
    None => format!("secretspec/{}/{}/{}", project, profile, key),
};
```

So `prefix=myteam` gives `myteam/secretspec/...`, which is correct, and `prefix=myteam/` gives
`myteam//secretspec/...`, which matches nothing.

## Reproduction

secretspec 0.18.0, eu-west-1.

1. Create the secret `myteam/secretspec/demo-project/dev/API_TOKEN`.
2. Read it with the correct prefix:

```sh
secretspec get API_TOKEN --provider "awssm://eu-west-1?prefix=myteam" --profile dev
# prints the value
```

3. Read it with a trailing slash:

```sh
secretspec get API_TOKEN --provider "awssm://eu-west-1?prefix=myteam/" --profile dev
# Error:   x Failed to get secret
#   `-> Secret 'API_TOKEN' not found
```

A trailing slash is an easy thing to write, because the documentation describes the prefix in terms
of the `myteam/` namespace and because IAM ARNs are written as `secret:myteam/*`.

## Suggested fix

Any of these would remove the trap:

1. Trim a trailing `/` from the prefix when parsing the URI, so both spellings work.
2. Reject a prefix with a trailing slash at parse time, with a message naming the problem.
3. Say explicitly in `docs/src/content/docs/providers/awssm.md` that the prefix takes no trailing
   slash, and show the resulting name.

Option 1 seems kindest. `awsps` has the mirror case, where the prefix is expected to start with `/`.

Happy to open a PR.

domenkozar:
## Summary

- add a shared provider-native path joiner that normalizes only the slash boundary between fragments
- use it for AWS Secrets Manager convention names and the equivalent Scaleway and Infisical joins
- cover trailing, repeated, root, leading, and internal slash behavior
- document the AWSSM fix in the changelog

## Root cause and impact

The AWSSM provider appended `/secretspec/...` directly to the configured prefix. A prefix already ending in `/` therefore produced `myteam//secretspec/...`, which made reads look missing and could direct writes to a distinct secret name.

The join now guarantees exactly one boundary slash while preserving leading and internal slashes, so `?prefix=myteam` and `?prefix=myteam/` resolve to the same documented name.

## Validation

- `devenv shell cargo test -p secretspec --lib` — 1260 passed, 4 ignored
- `devenv shell prek run -a` — clippy and rustfmt passed
- `git diff --check origin/main...HEAD`

Closes #344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - OnePassword optional references whose item names resemble authentication
   diagnostics are omitted as missing instead of aborting batch resolution.
 
+### Fixed
+
+- The `awssm` provider now accepts a trailing slash in `?prefix=` without
+  inserting a second slash into the AWS secret name. For example,
+  `?prefix=myteam/` resolves to `myteam/secretspec/...`, matching
+  `?prefix=myteam`, and both spellings share one provider identity so import
+  diagnostics still recognize alias-specific references. This avoids silently
+  treating the secret as missing or writing to a distinct double-slash name. Closes
+  [#344](https://github.com/cachix/secretspec/issues/344).
+
 ## [0.19.1] - 2026-08-11
 
 Republishes 0.19.0's command-line artifacts. The library and CLI behave exactly
--- a/secretspec/src/provider/awssm.rs
+++ b/secretspec/src/provider/awssm.rs
@@ -47,7 +47,7 @@
 //! secretspec check --provider awssm://production@us-east-1
 //! ```
 
-use super::{Address, Provider, ProviderUrl};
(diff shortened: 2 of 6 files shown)


moltar (approved):
Lgtm 

Thank you 🙏

domenkozar:
@copilot rebase this pull request

Issue #32, opened 2025-08-20
Support setting multi-line secrets


sandydoo:
Both the interactive and piped input methods read until the first newline character.
This makes it difficult to ingest certain multi-line secrets, like PEM encoded private keys.

### Interactive case

`rpassword` doesn't support mutli-line input. In fact, most password input libraries I've looked at don't.

We would also need a different way to detect EOF. The usual key combo appears to be `ctrl+d`, but maybe we could use enter/shift+enter for this?

There's also the question of what to display in the multi-line case. Echo should be disabled? Should be show that something was pasted in (e.g. like claude code does). Can this be done safely? Do we need to write a TUI for this?

### Piped input

We could read the entire input until EOF.

PR #351, opened 2026-08-15
refactor(ci): simplify release-pr workflow

ifiokjr:
The `release-pr` workflow duplicated its release-commit detection across its two jobs, wrote report files nothing reads, and carried a no-op concurrency setting — more moving parts than the two jobs need. The two jobs must keep working as-is: `create` refreshes the release PR on every push to `main`, and `tag` pushes release tags, creates the draft GitHub release, and comments on released issues when a release PR merges — `publish.yml` depends on both.

## Implementation

- The `create` job now exposes `is_release_commit` as a job output; the `tag` job consumes it instead of re-running the identical `monochange step release-record` detection on the same commit. The `create` job's detection is stricter (it checks both the commit subject and the release-record body marker, and captures stderr), so no detection capability is lost.
- The `tag` job's `if` is hoisted to the job level (`needs.create.outputs.is_release_commit == 'true'`), so the whole job — including devenv setup — is skipped on non-release pushes.
- Dropped the dead `tee`s to `/tmp/tag-report.json` and `/tmp/publish-release.json`; their consumer (the "trigger release workflow" step) was removed in #32.
- Removed `cancel-in-progress: false`, which is the GitHub Actions default.

_Created on behalf of Ifiok Jr. ([@ifiokjr](https://github.com/ifiokjr)) by pi using deepseek-v4-flash:0731 at xhigh thinking._


ifiokjr:
Closing: this PR was created in the wrong repository by an automation mistake. The intended PR is in ifiokjr/monosecret.

PR #352, opened 2026-08-15
fix(sdk): close() must attempt every as_path file

dngr2:
## The bug

`Resolved.close()` exists so secret-bearing temp files do not outlive the result. The **Python** and **Ruby** SDKs stop at the first file the OS refuses to remove, so every later secret stays on disk — the one outcome `close()` is there to prevent — and the caller has no way to learn which ones survived.

Demonstrated with three `as_path` secrets where the second cannot be removed:

```
before close:  [True, True, True]
close() raised PermissionError: [Errno 13] Permission denied: '/tmp/secret1_kxwezsvs'
after close:   [False, True, True]

secret files still on disk: 2 of 3
   /tmp/secret1_kxwezsvs  contents='super-secret-value'
   /tmp/secret2_0_99cy2z  contents='super-secret-value'
```

## Why this is a contract bug, not a preference

The Go and .NET SDKs already handle it — they record the first failure and attempt the rest:

| SDK | Behaviour |
|---|---|
| Go `secretspec.go:152` | `firstErr`, every file attempted |
| .NET `Models.cs:90` | `firstError`, throws after the loop |
| PHP `Resolved.php:72` | `@unlink`, loop continues |
| **Python** `__init__.py:142` | **aborts, strands the rest** |
| **Ruby** `secretspec.rb:90` | **aborts, strands the rest** |

.NET catches `IOException` specifically — the ordinary Windows sharing violation raised when another process still holds the file open. So a failed delete is already treated as an expected condition in this project; Python and Ruby simply do not implement that half of the contract. Both now follow it: clean up everything, then raise the first error.

## A second Ruby bug this closes

The Ruby guard was `File.delete(path) if File.exist?(path)`. `File.exist?` **follows symlinks**, so a dangling symlink reports absent and is never removed — it just accumulates in the temp dir. Rescuing `Errno::ENOENT` instead removes that case and the check-then-delete race together. Called out separately in the changelog rather than folded into the main entry.

## Verification

Both SDKs were built and tested against the real compiled extensions, not stubs.

- **Python** — `maturin develop --uv`; 5 new tests pass, existing suite 19 pass
- **Ruby** — native extension built via `scripts/build-ext.sh`; 4 new tests pass, existing `test_resolve.rb` 14 runs / 34 assertions / 0 failures

Mutation-tested: reverting either fix fails `test_close_removes_the_rest_when_one_file_cannot_be_removed` on exactly the intended assertion — *"file after the failure was stranded on disk"*. The other tests still pass under that mutation, so the stranding case is the one carrying the regression.

Tests cover the ordinary path, idempotency, that a refusal does not strand the other files, and that the first error is the one raised.

One note in the interest of full disclosure: `secretspec-py/tests/test_codegen.py::test_quicktype_types_consume_runtime_fields` fails in my environment with `ERR_REQUIRE_ESM` from `quicktype` under Node 18. I confirmed it fails identically with my changes stashed, so it is pre-existing and environmental, but I could not get that one test green locally.

CHANGELOG entry added under the existing `### Fixed` heading in `[Unreleased]`, per `AGENTS.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   spelled `n-u-l-l`. The `bw` and `dashlane` providers already behaved this way.
   An `extract` pointer is unchanged: it names one location and still reports a
   `null` there, and the two policies now sit next to each other in one place.
+- The Python and Ruby SDKs' `Resolved.close()`/`Resolved#close` now remove every
+  `as_path` temp file even when one of them cannot be removed, raising the first
+  such error only after the rest are cleaned up. Previously the first failure
+  aborted the loop and left the remaining secret files on disk, which is the
+  outcome `close` exists to prevent. This matches the Go SDK's `firstErr` and the
+  .NET SDK's `firstError`. The Ruby SDK also no longer skips a dangling symlink,
+  which `File.exist?` reports as absent.
 - The `awssm` provider now accepts a trailing slash in `?prefix=` without
   inserting a second slash into the AWS secret name. For example,
   `?prefix=myteam/` resolves to `myteam/secretspec/...`, matching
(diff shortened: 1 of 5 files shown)


domenkozar:

dngr2:
Rebased onto main. The only conflict was in CHANGELOG.md, between this entry and the one from #358; both are kept.

Re-ran both SDK suites on the new base: Python 19 passed, Ruby 18 runs across three files. `test_codegen` fails in my environment on main as well, from a quicktype/stream-json ESM mismatch with the local Node, so it is unrelated.

PR #357, opened 2026-08-16
feat(sdk): expose Secrets::config() publicly (hidden)

djbclark:
## Summary

`Secrets::config()` was `pub(crate)`, gated behind
`#[cfg(any(feature = "cli", test))]`, for `secretspec codegen`'s use. Any
other embedder needing the manifest without a provider — an out-of-tree
plugin inspecting declared secrets, for instance (#64) — had no way to reach
it short of taking the whole `cli` feature or re-parsing the manifest itself.

## Change

`config()` is now `#[doc(hidden)] pub` and unconditional. `#[doc(hidden)]`
keeps it off the public SDK surface (ordinary callers resolve secrets, not
manifests) while making it reachable. This is additive and cannot break the
C ABI the FFI bindings expose.

## Validation

- `cargo check -p secretspec --all-features`
- `cargo check -p secretspec --no-default-features`
- `cargo test --package secretspec --all-features -- secrets::` — 37 passed
- `cargo clippy -p secretspec --no-default-features --features cli -- -D
  warnings` — no new warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Secrets::config()` is now `#[doc(hidden)] pub` instead of `pub(crate)`
+  behind the `cli` feature. Any embedder that needs the manifest without a
+  provider — an out-of-tree plugin inspecting declared secrets, for
+  instance — can now reach it without taking the whole `cli` feature.
 - Structured caller context lets CLI and SDK integrations identify the invoking
   software, version, operation, and non-secret resource independently of the
   user-supplied access reason. Audit records and providers receive the context,
--- a/secretspec/src/secrets.rs
+++ b/secretspec/src/secrets.rs
@@ -1883,10 +1883,17 @@ impl Secrets {
         Ok(())
     }
 
-    /// Get a reference to the project configuration. Used by `secretspec
-    /// codegen` (which needs the manifest, not a provider) and by tests.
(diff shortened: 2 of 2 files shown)


domenkozar:
Needs to check it doesn't conflict with #334

djbclark:
Thanks for the pointer, @domenkozar — checked #334 against this directly rather than just against the diff.

Merging #334 into a branch with this PR is clean: `Secrets::config()` auto-merges with no conflict, only a trivial `CHANGELOG.md` text overlap. #334 does move `Config` behind `__private` though, so I also checked the actual blast radius against a downstream CLI I maintain that embeds `secretspec` as a library and calls `config()` for JSON Schema emission.

Turned out small. `secretspec-derive` is already updated by #334 itself — no work needed there. The one real change needed was swapping a `&Config`-based schema build for the public `Spec::load_from(...)` + `codegen::build_ir(&spec)` path, which is arguably cleaner than what it replaces. I also hit one genuine conflict of the kind you're asking about here: #334 renames the internal `build_ir` helper, and a test added by #355 still called the old signature — confirmed and fixed.

Diff of the concrete fix, if useful: https://github.com/frdminc/sudo-secretspec/commit/bf0b25c

`cargo test --all-features` passes clean with #334 + this PR + #355 + #356 all merged together.

domenkozar:
@djbclark I mean that the public api will be `Spec` and `Spec::from(path)` instead of the internal Config.

domenkozar:
closing in favor of https://github.com/cachix/secretspec/pull/334, please open an issue with your use case if that doesn't fit

Issue #345, opened 2026-08-14
Proposal: Generic `exec://` Provider Plugin Protocol (and an AI privilege boundary showcase)


djbclark:
Hi team! 👋 

We've been working on a solution to a growing security pain point in local development: **AI coding assistants running under the developer's user account.** 

Because AI agents run with the developer's permissions, they can easily bypass standard secret management procedures—accidentally reading `.env` files, dumping the local user keychain, or leaking tokens to external logs. To solve this, we built a mathematically verifiable privilege boundary called [`sudo-secretspec`](https://github.com/djbclark/sudo-secretspec). It stores secrets in a vault owned by a dedicated, isolated service user (e.g., `_secretspec`). This physical separation intentionally avoids using `root` while still completely preventing the AI from reading secrets directly, forcing it to request them through an audited `sudo` companion binary.

Our initial thought was to submit this macOS-specific dual-user daemon and installer as a PR to `secretspec`. However, we realized that adding OS-specific `sudo` wrappers directly into the core engine would add a heavy, ongoing maintenance burden to your project.

### The Proposal: An `exec://` Provider Plugin Protocol

Instead of asking you to merge OS-specific privilege code, we would love to propose implementing a generic **"External Exec Provider"** (similar to [Git credential helpers](https://git-scm.com/docs/gitcredentials), [Kubernetes exec plugins](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins), or [Terraform providers](https://developer.hashicorp.com/terraform/plugin)).

This would allow `secretspec` to delegate secret resolution to any arbitrary external binary using JSON over `stdin`/`stdout`.

**Example Configuration:**
```toml
[providers.macos_boundary]
# Or potentially configured via URI: exec:///usr/local/bin/sudo-secretspec
type = "exec"
command = ["/usr/local/bin/sudo-secretspec", "resolve"]
```

**How it would work:**
1. When `secretspec` resolves a secret mapped to this provider, it shells out to the `command`.
2. It pipes a standardized JSON payload to `stdin` (e.g., `{"action": "get", "address": {...}}`).
3. The external binary performs its custom logic (in our case, safely escalating to the service user via `sudo` and querying the isolated vault) and prints a JSON response to `stdout`.

### Why we think this is a win-win:
1. **Zero OS-specific maintenance**: `secretspec` remains completely agnostic to the underlying storage or privilege escalation mechanism.
[...]

PR #362, opened 2026-08-16
feat: add versioned client and provider IPC

### LINKED ISSUE #421: IPC v1: define callback limits during rpc.initialize
## Context

`client.credential` may run while `rpc.initialize` is active, but the selected limits do not exist until the initialize result is committed.

## Problem

`docs/src/content/docs/reference/ipc-wire.md` says:

- pre-negotiation permits one 1 MiB frame and one in-flight request;
- negotiated `max_in_flight` also bounds server-to-client callbacks;
- callbacks may be sent while `rpc.initialize` is still active.

It does not define which frame/in-flight limits govern those initialize-time callbacks, or whether limits are per direction.

Current reference behavior is asymmetric:

- `secretspec-ipc/src/client.rs` starts callback acceptance with `Limits::PRE_NEGOTIATION`, so it accepts one active callback and the absolute frame ceiling during initialization;
- `secretspec-ipc/src/server.rs::Peer` uses the absolute frame ceiling before negotiation but does not enforce an outbound callback in-flight count, so a handler can issue multiple concurrent callbacks that the reference client treats as a protocol violation.

This can make two individually plausible implementations incompatible or deadlock during authentication.

## Suggested normative rule

Define explicitly that, until the successful initialize response is committed:

1. the absolute 1 MiB frame ceiling applies in both directions;
2. at most one callback may be outstanding in the server-to-client direction;
3. the initialization request and callback occupy independent directional request spaces/limits;
4. after readiness, the selected `max_frame_bytes` and `max_in_flight` apply to callbacks;
5. the server must not send a callback that violates the active limit, and the client closes on a peer limit violation.

Alternatively, if the limits offered in initialize should govern callbacks immediately, state that and implement it consistently.

## Deterministic coverage

Add tests for one initialize-time callback, two concurrent initialize-time callbacks, an oversized initialize-time callback, transition to negotiated limits, and callback completion racing the initialize response.

### LINKED ISSUE #422: Make provider IPC conformance runnable against third-party endpoints
## Context

While validating an independent `secretspec.provider/1` endpoint, only the generic strict-wire case could be run meaningfully. The advertised provider conformance driver is coupled to the bundled memory endpoint.

## Current limitation

`conformance/ipc/runner/src/bin/ipc-provider-conformance-driver.rs` hardcodes assumptions including:

- provider URI `memory://conformance`;
- scheme/provider identity `memory`;
- semantic credential name `conformance_token` and scope `memory://conformance`;
- memory-provider convention-address behavior and mutable state.

`conformance/ipc/cases/provider-operations.json` also initializes scheme `memory`.

A conforming third-party endpoint with its own registered scheme and URI correctly rejects these assumptions, so failures do not distinguish protocol non-conformance from fixture inapplicability. This conflicts with the conformance README's goal of language-neutral cases usable by third-party provider endpoints.

## Suggested resolution

Provide one or both of:

1. a parameterized endpoint profile supplying executable/arguments, scheme, URI, supported capabilities, a disposable native address, credential request expectations, and safe setup/cleanup hooks; and
2. a transport/lifecycle-only provider profile that makes no provider-specific storage assumptions.

Keep the deterministic memory implementation as the full reference matrix, but make its provider semantics explicit rather than presenting those cases as generic endpoint conformance.

## Desired validation layers

A third-party endpoint should be able to run, without source changes:

- framing and malformed-wire rejection;
- initialize/version/limit negotiation;
- strict request and tolerant response evolution;
- capability gating;
- callback brokerage, IDs, parent deadlines, cancellation, and callback errors;
- shutdown/crash/reconnect lifecycle;
- optional provider operations selected only when the supplied profile declares safe fixtures.

The runner should clearly report `not applicable` for unsupported semantic cases rather than a protocol failure.

### LINKED ISSUE #423: IPC v1: enforce callback parent deadline and lifetime
## Context

The callback contract in `docs/src/content/docs/reference/ipc-wire.md` requires a callback to be a child of one active request:

- `_meta.parent_request_id` names the active parent;
- callback deadline must not be later than the parent deadline;
- cancelling the parent cancels the callback.

## Current gap

`secretspec-ipc/src/client.rs::serve_callback` verifies that `parent_request_id` exists in the pending request table, but it does not:

- compare the callback deadline with the parent's deadline;
- associate active callback tasks with parent cancellation/termination;
- stop a callback handler when the parent becomes terminal;
- explicitly zeroize a secret result produced after the parent ended.

The pending table currently stores only the response sender, not the parent deadline/cancellation state. A callback with a later deadline can therefore continue after the operation it claims to serve has been cancelled or completed.

## Suggested resolution

1. Store parent deadline and cancellation/terminal state with each pending outbound request.
2. Reject a callback whose parent is absent/terminal or whose deadline exceeds the parent deadline.
3. Give callback handlers a child cancellation token tied to parent cancellation, deadline, session close, and terminal response.
4. Ensure a late callback result is discarded and secret-bearing buffers are zeroized promptly.
5. Specify whether invalid parent metadata is answered with `invalid_request` before closing or is immediately transport-fatal.

## Deterministic coverage

Add cases for:

- missing and unknown parent IDs;
- callback deadline exactly equal to the parent deadline;
- callback deadline one millisecond beyond the parent;
- parent cancellation while the callback handler is running;
- parent terminal response racing callback completion;
- disconnect during callback;
- confirmation that a late credential value is never delivered or retained.

### LINKED ISSUE #424: IPC v1: specify and test connection closure after initialization-state violations
## Context

An independent provider implementation review found that correct pre-negotiation termination behavior had to be inferred from the Rust server rather than from an explicit state-machine rule.

## Problem

`docs/src/content/docs/reference/ipc-wire.md` says the receiver rejects:

- an application request before successful initialization;
- a second `rpc.initialize` request;
- unsupported initialization protocol/version combinations.

It does not consistently state whether the receiver sends one terminal error and keeps the connection open, or sends the error and closes it.

`secretspec-ipc/src/server.rs` currently sends `invalid_request` and closes for a pre-initialization application request or second initialize. During an active initialize, another inbound request also cancels initialization, receives `invalid_request`, and ends the session. This is security-relevant state-machine behavior and independent endpoints should not have to reverse-engineer it.

The checked-in `conformance/ipc/cases/wire-rejections.json` does not cover these valid-JSON state violations.

## Suggested normative rule

Define a failed initialization state explicitly. For example:

- before readiness, only the initial `rpc.initialize`, its cancellation, and responses to callbacks raised by that initialize are valid;
- any other request receives at most one value-free `invalid_request` response and the connection then closes;
- unsupported protocol/version returns `unsupported_version` and closes;
- a malformed/invalid initialize returns the appropriate error and closes;
- a second initialize after readiness returns `invalid_request` and closes;
- no application request is processed after any failed initialization.

Please also specify whether the client must close immediately when it receives an unmatched response before initialization.

## Deterministic coverage

Add black-box cases for:

- application request before initialize;
- response before initialize;
- second initialize while the first is active;
- second initialize after readiness;
- unsupported protocol/version;
- invalid initialization params;
- proof that the endpoint closes and performs cleanup after each case.

### LINKED ISSUE #425: IPC v1: resolve contradictory unknown-notification semantics
## Context

The notification evolution rules on `feat/ipc-v1` at `ce29ab8` / #362 currently point in different directions.

## Contradictions

In `docs/src/content/docs/reference/ipc-wire.md`:

- the JSON-RPC profile says unknown notifications are ignored;
- the forward-compatibility section says an unknown notification method remains a protocol violation;
- the same section says request and notification objects reject unknown members.

The artifacts differ too:

- `schema/ipc/v1/common.schema.json` constrains `method` to `rpc.cancel` but sets `additionalProperties: true` on `NotificationEnvelope`;
- `secretspec-ipc/src/jsonrpc.rs::Notification` uses `#[serde(deny_unknown_fields)]`;
- `secretspec-ipc/src/client.rs` and `server.rs` effectively ignore unknown notification methods after envelope parsing.

This leaves independent implementations unable to know whether an unknown notification should be ignored, produce diagnostics, or terminate the session.

## Suggested resolution

Prefer normal JSON-RPC notification semantics:

1. Keep the notification envelope structurally strict (`jsonrpc`, `method`, `params`; no unknown top-level members).
2. Ignore an unknown notification method because it has no response channel.
3. Continue ignoring unknown/completed cancellation IDs and malformed `rpc.cancel` parameters as already documented.
4. Remove the statement that unknown notification methods are protocol-fatal, or explicitly choose the opposite behavior and make code/schema agree.
5. Explain how future notification methods are capability-gated even when older receivers ignore them.

## Deterministic coverage

Add conformance cases for:

- unknown notification method;
- unknown top-level notification member;
- malformed `rpc.cancel` params;
- cancellation for unknown and terminal IDs;
- a valid request after each ignored notification case to prove whether the session remains usable.

domenkozar:
## Summary

This PR adds two versioned, private-child-process IPC protocols for SecretSpec:

| Direction | Protocol | Purpose |
| --- | --- | --- |
| Application / SDK → SecretSpec | `secretspec.resolver/1` | Resolve, reject, release, and optionally mutate secrets through `secretspec serve` |
| SecretSpec → external provider | `secretspec.provider/1` | Implement provider operations in an independently shipped executable |

Both protocols use strict JSON-RPC over bounded newline-delimited stdio frames. They share initialization, version and capability negotiation, limits, deadlines, cancellation, structured errors, and orderly shutdown, while keeping their method sets separate.

This is deliberately not a daemon or socket protocol: every session launches a directly connected private child process without a shell.

## What this PR adds

- A new `secretspec-ipc` Rust crate with async client/server primitives, protocol types, lifecycle management, and a blocking resolver client.
- A new `secretspec-ipc-conformance` crate that exercises Rust and C implementations against shared malformed-input and lifecycle cases.
- `secretspec serve` and `secretspec serve --read-only` as the resolver endpoint used by applications and SDKs.
- External provider discovery, validation, launch, capability negotiation, and adaptation into the existing provider trait.
- A dependency-free C11 resolver client in `libsecretspec-resolver`, with CMake and Meson support and system-provided yyjson.
- Prompt callbacks for interactive resolution, including a polling API suitable for C and foreign-language runtimes.
- Reference documentation for the architecture, shared wire format, both protocols, and implementation guidance.
- A protocol-freeze gate and broader cross-platform CI coverage, including standalone C tests on Windows.

## Resolver protocol

Applications launch `secretspec serve` directly and select `secretspec.resolver/1` during initialization. Initialization fixes the manifest source, base directory, profile, provider selection, scope, and reason for the lifetime of the session.

Core methods:

- `resolver.get` requests exactly one named secret and returns either an inline value or a leased file.
- `resolver.release` releases leased material; outstanding leases are also cleaned up on disconnect.
- `resolver.reject` invalidates only the derived or cached copy that the caller rejected. It remains available in read-only sessions and is required in the 0.20+ protocol surface.
- `resolver.set` and `resolver.delete` are optional 0.20+ capabilities advertised by writable endpoints.
- `client.prompt` lets the resolver request user input when the client advertises prompt support.

`secretspec serve --read-only` does not advertise mutation methods and rejects resolution paths that would need to persist generated or prompted values. Headless clients receive an immediate missing/error result instead of leaving an unanswered prompt pending.

The C client exposes pending prompts through take/answer/decline/wait operations rather than invoking callbacks into an unknown foreign runtime. Prompt answers are validated before they are consumed, including strict UTF-8 validation.

## External provider protocol

External providers are registered as trusted executable definitions with an absolute executable path, fixed arguments, and runtime credential brokerage. SecretSpec launches one private child per provider URI and reason, then selects `secretspec.provider/1` during initialization.

The protocol covers provider naming and addressing, `get`, `get_many`, `exists`, `set`, expiring writes, `delete`, `clear`, writability checks, descriptions, preflight checks, and reflection. Capabilities are negotiated independently, so narrow or write-only providers do not need to pretend to support unrelated operations.

Provider failures retain structured error information. Session loss and reconnect behavior distinguish replay-safe requests from mutations that must not be replayed automatically.

## Library boundaries

- `secretspec-ipc` is the Rust implementation of both IPC protocols.
- `libsecretspec-resolver` is the pure-C11 client for `secretspec.resolver/1` only.
- `libsecretspec` remains the embedded, in-process resolver ABI used by existing SDK integrations. It was renamed from `secretspec-ffi`; the three exported C symbols remain unchanged, and loader compatibility is documented and tested.

Keeping the embedded ABI separate from the child-process protocol makes the security and lifecycle model explicit for each integration style.

## Security and lifecycle hardening

- Strict message shape, frame-size, nesting, string-size, ID, and deadline limits are enforced before dispatch.
- Request deadlines have a bounded horizon and propagate through callbacks; cancellation frees in-flight capacity without making a healthy session unusable.
- Expired or cancelled request and callback IDs are retained long enough to consume one terminal response safely.
- Child exit handling drains buffered responses before declaring the transport closed, including endpoints that reply and exit immediately.
- Startup failures kill and reap children with a fresh grace period, avoiding unreaped processes after an expired startup deadline.
- External executables are canonicalized and validated before launch. Windows validation walks the complete canonical ancestor chain and checks relevant ACL rights, including subtree replacement through `FILE_DELETE_CHILD`.
- Windows launches use constrained handle inheritance and a case-insensitively sorted Unicode environment block after inherited values and overrides are merged.
- Unix launch validation and permission checks prevent unsafe executable and directory configurations.
- The native clients reject malformed UTF-8, truncated frames, invalid JSON-RPC envelopes, duplicate keys, unknown IDs, oversized values, and unsupported batches.

## Conformance and testing

The conformance suite now runs a shared malformed-input corpus across the Rust resolver client, C resolver client, provider endpoint, and external-provider adapter. It covers empty and batch messages, duplicate keys, invalid UTF-8, truncated headers and payloads, oversized frames, and unknown response IDs.

Lifecycle coverage includes:

- repeated immediate-exit-after-shutdown-response sessions;
- a real Linux `/proc` assertion that startup-timeout children are reaped;
- end-to-end expired callback responses without session loss;
- in-flight slot reuse after cancellation, deadline expiry, and dropped requests;
- inherited and replaced Windows environment-variable ordering.

Focused gates run during the hardening pass:

```text
cargo test -p secretspec-ipc
cargo test -p secretspec-ipc-conformance
cargo test -p secretspec --test ipc_resolver
cargo check -p secretspec-ipc --tests --target x86_64-pc-windows-gnu
cmake -S libsecretspec-resolver -B <build-dir>
cmake --build <build-dir>
ctest --test-dir <build-dir> --output-on-failure
npm --prefix docs run build
```

The full Linux, macOS, Windows, SDK, packaging, feature, and documentation matrix also runs in CI.

## Documentation

- [IPC architecture](https://feat-ipc-v1-secretspec.domen.workers.dev/reference/ipc-architecture/)
- [Shared wire format](https://feat-ipc-v1-secretspec.domen.workers.dev/reference/ipc-wire/)
- [Resolver protocol](https://feat-ipc-v1-secretspec.domen.workers.dev/reference/resolver-protocol/)
- [Provider protocol](https://feat-ipc-v1-secretspec.domen.workers.dev/reference/provider-protocol/)
- [Implementation guide](https://feat-ipc-v1-secretspec.domen.workers.dev/development/ipc-implementation/)

## PR #98 feedback incorporated

This design preserves the useful direction from #98—version negotiation, capability discovery, explicit limits, structured errors, preflight checks, and external-provider extensibility—while tightening process ownership, least-access resolution, cancellation/deadline behavior, non-replayable mutations, lease cleanup, prompt handling, executable trust validation, and cross-language conformance.

## Follow-up protocol hardening

- Enforce initialize-time callback concurrency and size limits, including the pre-negotiation bootstrap limit.
- Bind callbacks to live parent requests and their effective deadlines, cancelling handlers and zeroizing discarded secret-bearing results.
- Add third-party transport-only provider profiles with explicit structured `not_applicable` results for semantic cases.
- Specify failed-initialization closure and align Rust/C handling of malformed, unknown, cancellation, and terminal notifications.
- Expand Rust, C, and conformance regression coverage for all of the above.

Fixes #421
Fixes #422
Fixes #423
Fixes #424
Fixes #425


djbclark:
Following up from #345 (closed into #64) — `secretspec.provider/1` is the out-of-tree provider mechanism that proposal asked for, so: thank you, and we intend to be an early adopter. [sudo-secretspec](https://github.com/frdminc/sudo-secretspec) plans to ship a System-scope registration whose endpoint is a thin unprivileged shim crossing a sudo-mediated privilege boundary into a root-owned broker — the "privileged integrations construct an allowlisted environment" case the architecture doc already nods at. As far as we can tell from the PR, that needs no upstream patching: `ProviderHandler`/`ProviderApplication` are public API and registration is data.

One piece of review feedback from operating that kind of boundary, aimed at the unix registration trust checks in `external.rs` (as of 337950c). Offered as hardening plus conformance cases, not an objection to the design:

1. **macOS ACLs are invisible to the mode-bit check.** `check_file_security`/`check_parent_security` gate on `metadata.mode() & 0o022`, but on macOS an extended ACL can grant `add_file`/`write` to a group while the POSIX bits read `0755` — third-party installers add such ACLs routinely. The Windows path already validates ACLs (`path_acl_is_trusted`); the unix path currently has no ACL story. sudo-secretspec's drift checker refuses any extended ACL on a protected path or its ancestors for this reason.

2. **Both checks follow symlinks.** `std::fs::metadata` resolves, so what gets validated is the resolved target, not the literal registered path. A symlinked component anywhere in the chain silently redirects discovery to wherever a root-owned, mode-clean target happens to be. `symlink_metadata` plus an explicit no-symlinked-components rule is cheap at discovery time.

3. **Trust stops at the immediate parent.** Everything above the registration directory's parent is implicitly trusted. On stock macOS 26 the default System chain (`/Library/Application Support`, `root:admin 0755`, no ACL — verified) is sound as shipped, but that soundness is assumed rather than checked, and one loosened ancestor — by mode or by ACL, see (1) — defeats both checks below it. Walking the resolved chain to `/` once per discovery is what sudo-secretspec does (`check_ancestor_chain` in its drift checker): refuse non-root-owned, group/world-writable, ACL-bearing, or symlinked ancestors.

Happy to turn this into (a) a small hardening PR against `feat/ipc-v1` and (b) conformance-suite cases — ancestor-writable-by-mode, ancestor-writable-by-ACL, symlinked-component, symlinked-registration-file — if that would be useful. And once the privileged endpoint ships against this we'll write the deployment up so the reference case is documented.

djbclark:
### macOS: `libsecretspec-ipc` does not build — `sigtimedwait` is not available on Darwin

Following up on [my earlier comment](https://github.com/cachix/secretspec/pull/362#issuecomment-5316998149): I built the first out-of-tree **privileged** `secretspec.provider/1` endpoint against `feat/ipc-v1` at `a393a27` this week. The endpoint side went cleanly — three methods, no patching of this PR, and it passes the checked-in `conformance/ipc/cases/provider-lifecycle.json`. Full write-up in a moment; one portability bug is worth reporting on its own first, because it blocks the conformance suite entirely on macOS.

`libsecretspec-ipc/src/process_posix.c:237` calls `sigtimedwait` under a plain `#ifndef _WIN32`:

```c
if (count <= 0) {
    if (mask_changed && !sigismember(&previous, SIGPIPE)) {
        struct timespec zero = {0, 0};
        (void)sigtimedwait(&blocked, NULL, &zero);
    }
```

**macOS has no `sigtimedwait` at all.** It is not a header-guard or feature-test-macro problem, and not a strict-mode warning:

```console
$ grep -rn sigtimedwait "$(xcrun --show-sdk-path)/usr/include/"   # no matches
$ nm -g /usr/lib/libSystem.B.dylib | grep -i sigtimedwait          # no matches
```

So it is a link failure as well as a compile error. With `-std=c11 -Werror` it surfaces at compile time as:

```
process_posix.c:237:23: error: call to undeclared function 'sigtimedwait';
ISO C99 and later do not support implicit function declarations
```

Practical effect: `cargo build -p secretspec-ipc-conformance` fails on macOS, because that crate's `build.rs` compiles the pure-C client. That takes out `ipc-provider-conformance-driver`, `ipc-client-conformance-driver`, and the `provider_cases` / `client_cases` tests — i.e. a third-party endpoint author on macOS cannot run the conformance suite this PR ships for them. (macOS 26.6, Apple clang, aarch64.)

#### Suggested fix

`sigpending` + `sigwait` is the portable equivalent, and it cannot block here: SIGPIPE generated by this thread's own `write` is thread-directed, so if it is pending it stays pending until this thread accepts it.

```diff
         if (count <= 0) {
             if (mask_changed && !sigismember(&previous, SIGPIPE)) {
-                struct timespec zero = {0, 0};
-                (void)sigtimedwait(&blocked, NULL, &zero);
+                sigset_t pending;
+                int signal_number;
+                if (sigpending(&pending) == 0 && sigismember(&pending, SIGPIPE)) {
+                    (void)sigwait(&blocked, &signal_number);
+                }
             }
             if (mask_changed) (void)pthread_sigmask(SIG_SETMASK, &previous, NULL);
             return false;
```

With that applied the conformance driver builds and runs on macOS unchanged. Two notes on the shape:

- `sigpending` reports the union of thread-pending and process-pending signals, so in principle this could consume a SIGPIPE generated elsewhere in the process. That is strictly better than the status quo of not building, and if you would rather keep `sigtimedwait` where it exists, an `#if defined(__APPLE__)` split around just these lines is equally fine by me.
- I deliberately did **not** apply this in my prototype tree — I kept it as a patch file — so that "an out-of-tree privileged endpoint needs no patching of upstream" stayed literally true and mechanically checkable (`git diff` empty). Happy to open it as a PR against `feat/ipc-v1` if that is easier than a review comment.

Related, since it is the same lane: #377 (SIGPIPE default disposition) is still open from my side.

domenkozar:
The INI `extract` support has been split out into #386 and removed from this branch.

It was unrelated to the IPC work and only shared a commit with it, so the two can now be reviewed and merged independently. This branch is unchanged otherwise: the first commit was rewritten to drop the INI hunks and the remaining 18 were replayed on top, so the IPC diff is byte-identical to what it was.

---
Created by Claude Code (see https://github.com/cli/cli/issues/13904 for why this note is here rather than in the GitHub API).

domenkozar:
Removed the Factorseal provider from this branch. SecretSpec no longer links the `factorseal` crate at all.

The provider was compiled in against `factorseal`'s `agent-client` feature through a `path = "../factorseal"` dependency on an unpublished crate, and it was in the default feature set. That made `cargo build` fail for anyone without a sibling checkout, and it would have blocked `cargo publish` for 0.20.

The right shape is for Factorseal to ship a `secretspec.provider/1` endpoint and for SecretSpec to reach it through the external-provider adapter this PR adds. That inverts the dependency to factorseal -> `secretspec-ipc` and puts every Factorseal-specific concern on Factorseal's side of the protocol: the convention address and cache namespace become `provider.resolve_address`, the agent socket discovery becomes the endpoint's business, and the error translation becomes the protocol's structured error kinds. Nothing Factorseal-specific is left for SecretSpec to hold, so the module is gone rather than rewritten. A scheme with no built-in registration already falls through to external discovery, so `factorseal://` needs no new machinery when the endpoint exists.

Also dropped: the provider doc page, the README entries, the changelog entry, and the `directories` dependency. The protocol docs that stated Factorseal is in-tree and does not use the external protocol have been corrected.

The branch is now rebased onto current main, squashed to one commit, and builds and tests with default features: 1528 passing, 0 failing.

---
Created by Claude Code (see https://github.com/cli/cli/issues/13904 for why this note is here rather than in the GitHub API).

domenkozar:
One use case we ought to support is IPC via SSH or some remote protocol.

We might want to keep the vault local but remote session to access some stuff.

PR #365, opened 2026-08-16
fix(node): release AWS state before process exit

### LINKED ISSUE #343: Node SDK: process never exits after reportAsync or loadAsync with the AWS providers
## Summary

The Node SDK can finish its work and then leave the Node process alive forever. `reportAsync()`
returns a valid report, or `loadAsync()` returns valid values, and the event loop never drains.

It only happens with the AWS providers. The same calls against `env://` always exit cleanly, so
this looks like an AWS client or tokio runtime that is not dropped rather than a problem in the
async binding itself.

Any short lived process is affected. A Lambda, a CI step or a migration script burns its whole
timeout instead of returning.

## Versions

- `secretspec` npm 0.18.0 and 0.19.0, both affected
- Node 24.19.0, darwin-arm64
- Region eu-west-1

## Observed behaviour

| Call | Result |
| --- | --- |
| `reportAsync()` on some profiles | returns the report, process never exits |
| `loadAsync()` that rejects with `MissingRequiredError` | rejection is delivered, process never exits |
| `loadAsync()` on some profiles | returns the values, process never exits |
| `loadAsync()` / `reportAsync()` on other profiles | exit cleanly |
| `load()` and `report()`, synchronous | always exit cleanly |
| any of the above against `env://` | always exit cleanly |

It is deterministic, not intermittent. Each case above was repeated five times with the same result.
A given manifest and profile pair either always hangs or never does.

## Reproduction

Manifest, with two AWS secrets that exist:

```toml
[project]
name = "demo-project"
revision = "1.0"

[providers]
sm = "awssm://eu-west-1"
ps = "awsps://eu-west-1"

[profiles.smonly.defaults]
providers = ["sm"]
required = true

[profiles.smonly]
A = { description = "sm", ref = { item = "demo/legacy-token" } }

[profiles.psonly.defaults]
providers = ["ps"]
required = true

[profiles.psonly]
B = { description = "ps", ref = { item = "/demo/parameter" } }
```

Script:

```js
const { SecretSpec } = require('secretspec');

(async () => {
  const r = await SecretSpec.builder()
    .withPath(process.argv[2])
    .withProfile(process.argv[3])
    .withReason('repro')
    .loadAsync();
  console.log('resolved', r.provider);
  r.dispose();
  console.log('end of script');
})();
```

Run it:

```sh
timeout 30 node repro.js ./secretspec.toml smonly
echo "exit=$?"
```

Output:

```
resolved awssm://eu-west-1
end of script
exit=124
```

The script reaches its last line, then the process hangs until `timeout` kills it. Replacing
`loadAsync()` with `load()` gives `exit=0`. The same script against a manifest using `env://` gives
`exit=0`.

`reportAsync()` reproduces it the same way, including on a profile whose required secret is missing,
where it correctly returns `status: "missing_required"` and then hangs.

## What I could not narrow down

I could not isolate the exact trigger. Some profiles hang every time and others never do, with no
pattern I could pin from the outside. Things I ruled out:

- Number of secrets in the profile. One-secret and two-secret profiles both hang and both pass,
  depending on the manifest.
- Declaring a provider alias that no profile uses. Adding an unused alias did not cause a hang.
- Repeated calls. Three sequential `loadAsync()` calls on a passing profile still exit cleanly.

What held in every test: the AWS providers are involved whenever it hangs, `env://` never hangs,
and the synchronous entry points never hang.

`dispose()` is called in the repro, so it is not leaked `as_path` temp files.

## Workaround

Use `load()` and `report()`, or call `process.exit()` once the work is done.

domenkozar:
## Summary

- run each N-API async provider operation on a scoped Rust thread
- destroy provider thread-local runtime state before the libuv worker returns
- add natural child-process exit regressions for AWS Secrets Manager and Parameter Store using a local compatible server

Fixes #343

## Testing

- devenv shell npm --prefix secretspec-node test (17 of 17 passed)
- fresh isolated-branch AWS exit regression (2 of 2 passed)
- devenv shell cargo clippy -p secretspec-node-native --no-deps -- -D warnings

## Note

The original report is macOS-specific. Local validation was on Linux, so reporter confirmation on macOS remains useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   keeping the server pinned independently of ambient `bws` configuration.
   ([#359](https://github.com/cachix/secretspec/issues/359))
 
+- Node SDK processes using `loadAsync()` or `reportAsync()` with AWS Secrets
+  Manager or Parameter Store now exit normally after resolution. Provider
+  runtime and TLS state is torn down on a short-lived resolver thread instead
+  of remaining attached to a persistent libuv worker during macOS process
+  shutdown. ([#343])
+
+  [#343]: https://github.com/cachix/secretspec/issues/343
+
 - The `awssm` and `scaleway` providers now treat a JSON `null` in a `ref` field
   as no value, the same as an absent key, so the provider chain continues.
   Previously it was rendered as the four-character string `null`, which
--- a/secretspec-node/src/lib.rs
+++ b/secretspec-node/src/lib.rs
@@ -18,8 +18,9 @@ pub fn resolve(request_json: String) -> String {
     secretspec::resolve_json(&request_json)
 }
 
-/// Runs `resolve_json` on the libuv threadpool (via [`AsyncTask`]) so it never
-/// runs on the JS thread.
+/// Dispatches `resolve_json` from the libuv threadpool to one short-lived Rust
+/// thread, so it never runs on the JS thread and provider runtime/TLS state is
+/// torn down before the libuv worker is returned to Node.
 pub struct ResolveTask {
     request_json: String,
 }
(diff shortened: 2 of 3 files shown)


moltar:
Confirmed on macOS (26.5.2, arm64, Node 24.19.0): built from source at cf48a7a, the exit hang is resolved for both awssm and awsps. Details, an offline HTTPS repro, and stack traces from a hung 0.19.1 process are in https://github.com/cachix/secretspec/issues/343#issuecomment-5325877233. One note: the new plain-HTTP regression test passes on the broken 0.19.1 as well, because the trigger is a TLS connection attempt. The linked comment has a TLS harness you can lift for the test.

PR #367, opened 2026-08-16
fix(gcsm): make convention names collision-safe

### LINKED ISSUE #219: Secrets from two projects can collide on one stored name
Two different `(project, profile, key)` triples can name one stored secret: `project` or `profile` can contain the delimiter a provider joins with (`is_valid_identifier` runs only on secret names). With `gcsm`, which joins with `-`, that needs only ordinary names: `("my-app", "prod", "K")` and `("my", "app-prod", "K")` both address `secretspec-my-app-prod-K`.

Injectivity seems to be the goal: distinct triples always produce distinct names. This would happen if the delimiter is absent from `project`, `profile`, and `key`.

Two approaches to achieve this:

- `akv` is a working example: it Base32-encodes each component and then joins with `--`, e.g. `("my-app", "prod", "K")` -> `secretspec--nv4s2ylqoa--obzg6za--jm`. Under this (or a similar) approach, consumers would have to re-store each existing secret under its new identifier: read old name, write new name, delete old.
- Validating `project` and `profile` like secret names. This would need `init` to stop seeding the name straight from the directory, since it rejects a hyphen. Existing configs using one would fail to load, and fixing them means renaming the project, changing the identifier as well, so another re-store exercise for users.

I lean towards the first option: it seems like a less onerous issue for users to deal with, while also limiting the code change to where it would be expected to be and without modifying other existing behaviours.

What do you think?

domenkozar:
## Summary

- replace ambiguous hyphen flattening with reversible lowercase Base32 components
- add an exact collision regression and property tests
- document the 0.20 naming migration; explicit ref addresses remain unchanged

Fixes #219

## Testing

- devenv shell cargo test -p secretspec gcsm
- devenv shell cargo test -p secretspec
- devenv shell npm --prefix docs run build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Google Cloud Secret Manager convention names now encode the project,
+  profile, and key separately with lowercase, unpadded Base32. Distinct
+  logical addresses such as `my-app/prod/K` and `my/app-prod/K` can no longer
+  collide on one stored secret. Convention secrets created by SecretSpec 0.19
+  or earlier keep their old GCSM ids and must be re-stored for the 0.20 naming
+  convention; explicit `ref` addresses remain unchanged. ([#219])
+
+  [#219]: https://github.com/cachix/secretspec/issues/219
+
 - Bare `bws://<project-uuid>` provider URIs now target the Bitwarden US cloud
   vault instead of the public marketing site, restoring reads and writes while
   keeping the server pinned independently of ambient `bws` configuration.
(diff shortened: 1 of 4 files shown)


domenkozar:
Superseded by #368, which keeps readable versioned GCSM names and migrates legacy entries lazily on the first convention read.

ojhermann:
Upgrading renames every existing gcsm convention secret, so they all read as unset. Users will re-enter the values, which leaves the originals live and unreferenced in GCP, and gcsm implements no `delete`, so nothing can clean them up.

Probing the legacy id when a convention read misses would report the rename instead of a silent "missing", and a migration note could document an alias-level `ref` template of the old name, which makes the originals addressable so `import` can copy the values across. Happy to put that up as a PR.

PR #373, opened 2026-08-17
fix(check): write the report to stdout so it can be piped

### LINKED ISSUE #372: `check` writes its entire report to stderr, so `secretspec check | grep ...` silently returns nothing
### ELI5

`secretspec check` prints a nice report — a header, a ✓/○/✗ line per secret, and a
summary. It looks like normal program output. But all of it is going out the
"error" pipe instead of the "normal" pipe.

So if you try to do the obvious thing:

```console
$ secretspec check | grep DATABASE_URL
```

…you get nothing, and it looks like `check` found no secrets. It found them
fine — the text just went somewhere `grep` wasn't looking.

### The bug

Every line `check` emits goes to stderr. stdout is completely empty. Captured
against `main` at `dfa4b10`, with a two-secret manifest and a `dotenv` provider:

```console
$ secretspec check --reason "deploy preflight" 2>/dev/null   # stdout only
[exit 0]

$ secretspec check --reason "deploy preflight" | grep DATABASE_URL
[grep exit 1]
```

Both silent. The report is there — it's on stderr — so the exit status says
"all good" while every pipeline sees an empty report.

Source, on `main` at `dfa4b10` (`secretspec/src/secrets.rs`):

- `check()` — the header, line 3678
- `display_validation_success()` — every status line and the summary, 3701
- `display_validation_errors()` — same for the failure path, plus constraint
  violations, 3731

11 `eprintln!` calls across the three.

With them writing to stdout instead, same manifest, same commands:

```console
$ secretspec check --reason "deploy preflight" 2>/dev/null   # stdout only
Checking secrets in demo (profile: default)...

✓ DATABASE_URL - app database
○ SENTRY_DSN - error reporting (optional)

Summary: 1 found, 0 missing, 1 optional
[exit 0]

$ secretspec check --reason "deploy preflight" | grep DATABASE_URL
✓ DATABASE_URL - app database
[grep exit 0]
```

### Why this is worth fixing beyond the piping annoyance

The colour handling is currently self-contradictory. `colored` decides whether
to emit ANSI escapes by testing **stdout**:

```rust
// colored-3.1.1/src/control.rs:108
&& io::stdout().is_terminal(),
```

But the report is written to **stderr**. The two streams are independent, so
both mistakes are reachable today:

- `secretspec check 2>log` with a terminal stdout → `colored` sees a TTY and
  emits escapes, which land as raw `\x1b[32m` bytes **inside the log file**.
- `secretspec check >file` with a terminal stderr → `colored` sees a pipe and
  strips colour from output the user is actually reading on their terminal.

Routing the report to stdout makes the existing TTY detection correct as a side
effect, rather than requiring any new colour logic.

### `check` already contradicts itself

The clearest evidence that this is an oversight rather than a decision is that
the *same subcommand* routes the *same report* to different streams depending
on a flag:

- `check --json` → `println!` (`cli/mod.rs:1507`) — stdout
- `check --explain` → `print!` (`cli/mod.rs:1509`) — stdout
- `check` → `eprintln!` (`secrets.rs:3678` and the two `display_validation_*`
  helpers) — stderr

So a caller can already pipe the machine-readable and the explain renderings,
but not the default human one. Whatever the right stream is, it should not
depend on which flag was passed.

More broadly, a command's primary output — the thing the user asked for — goes
to stdout, and stderr carries diagnostics, progress and errors. `export` writes
to stdout via `io::stdout()` (`cli/mod.rs:1466`), as do `get`, `schema` and
`completions`.

**In fairness, `check` is not the only stderr reporter.** `import`'s summary is
also on stderr (`secrets.rs:4313`, `:4320`, `:4328`), as are the interactive
set/generation confirmations. I've left `import` alone here because I don't want
to widen a focused fix, but if you consider its summary primary output too, I'm
happy to include it — or to be told the house rule is the opposite of what I've
assumed, in which case the consistent fix is to move `--json`/`--explain` to
stderr instead and I'll send that patch.

### Proposed fix

`eprintln!` → stdout at the 11 call sites across those three functions. No
change to exit codes, no change to what is printed, no new dependency, no
colour rework. `ensure_secrets` is deliberately untouched — its output is
prompts and diagnostics, not report. `check --json` and `--explain` return
early at `cli/mod.rs:1443-1465`, before the header, so neither is affected.

**One thing worth doing at the same time, and the reason I'd not just swap the
macros.** Putting the report on stdout puts it on a stream a reader can close
early, and `println!` panics on EPIPE. Since Rust's stdout is a `LineWriter`,
each report line is its own write, so this needs no large output:

```console
$ secretspec check --reason r | head -1
Checking secrets in demo (profile: default)...
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
[exit 101]
```

That pipeline exits 0 today, precisely because stdout gets nothing. So a naive
macro swap trades a silent-empty-report bug for a panic on `| head`.

My patch therefore writes the report through an injected sink, which is the
pattern the codebase already documents for this exact reason —
`write_export`'s doc comment says an injected sink "turns a broken pipe into a
returned error instead of a panic". `check` locks stdout once and passes
`&mut dyn io::Write` to both display helpers. A closed pipe then behaves as
`export | head` already does: a clean `IO error: Broken pipe`, exit 1.

Two related things I noticed while doing this, both pre-existing and neither
touched by my patch, flagging in case you want them: `check --json | head`
already panics the same way (`cli/mod.rs:1456`), and if you'd rather a broken
pipe exit 0 silently — as `git` and `ls` do — that's a different and slightly
larger change, since `export` would presumably want to match. Happy either way;
tell me which and I'll do it consistently across both.

Open question for maintainers, happy to go either way:

- The constraint-violation lines in `display_validation_errors` are arguably
  diagnostics rather than report. I'd keep them on **stdout** with the rest of
  the report, on the grounds that splitting one human-readable report across two
  streams is the same class of bug — but say the word and I'll send them to
  stderr.

### Use cases this unblocks

1. **CI / automation.** `secretspec check | grep -q MISSING_KEY` and friends
   currently misreport. Today every wrapper has to know to add `2>&1`, which is
   non-obvious precisely because the command looks like it is printing normally.
2. **Logging with colour.** Capturing the report to a file while keeping a clean
   terminal is currently impossible without escape codes leaking into the file
   (see above).

### Library API

`Secrets::check()` is public (`secrets.rs:3674`) and appears in the SDK doc
examples, so this makes a *library* write to its consumer's stdout — a fair
objection, and arguably worse than writing to stderr, since a consumer emitting
JSON on stdout would have the report interleaved into it.

Two things make me think it's still right as proposed. The printing already
happens unconditionally today, just to the other stream, so no consumer is
currently spared it; and in practice the CLI is the only caller — the one
non-CLI caller I know of is my own downstream broker.

That said, if you'd rather not have a library print at all, the cleaner shape is
to move rendering into the CLI layer, or have `check` take an `impl Write`
(defaulting to `io::stdout()`) so an embedder can redirect it. I'm happy to do
either instead; say which and I'll send that patch.

### Compatibility

Anyone capturing via `2>&1` keeps working — that captures both streams. The
breaking case is a script capturing stderr *specifically* (`2>report.txt` with
stdout discarded).

I won't claim that's rare, because it caught me: my own downstream post-install
suite asserted the report was on stderr and asserted stdout was empty, and I had
to update three fixtures. Worth a `Changed` note rather than only a `Fixed` one,
and I'd expect a small number of wrappers to need the same edit.

I have a patch with tests and can open a PR immediately if the approach looks
right.

djbclark:
Fixes #372, where the reasoning and the transcripts are laid out in full.

`check` emitted its entire report through `eprintln!`, so stdout was empty and
`secretspec check | grep DATABASE_URL` matched nothing while the command exited
0 — which reads as "no secrets found" rather than "the report went elsewhere".

Three things beyond the piping annoyance, in the order I'd weigh them:

- **`check` already contradicts itself.** `check --json` (`cli/mod.rs:1507`) and
  `check --explain` (`:1509`) write to stdout. Only the default human rendering
  did not, so which stream carried the report depended on a flag.
- **The colour handling is currently wrong in both directions.** `colored`
  decides on ANSI escapes by testing *stdout* (`control.rs:108`) while the
  report went to *stderr*. So `check 2>log` leaked raw escape bytes into the log
  file, and `check >file` stripped colour from output being read on a terminal.
  Routing the report to stdout makes the existing detection correct rather than
  needing new colour logic.
- **A naive macro swap introduces a panic.** Putting the report on stdout puts
  it on a stream a reader can close early, and `println!` panics on EPIPE.
  stdout is a `LineWriter`, so every report line is its own write syscall and no
  large output is needed:

  ```console
  $ secretspec check --reason r | head -1
  Checking secrets in demo (profile: default)...
  thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
  failed printing to stdout: Broken pipe (os error 32)
  [exit 101]
  ```

  That pipeline exits 0 today only because stdout receives nothing. So this
  writes the report through a locked `&mut dyn io::Write` instead — the pattern
  `write_export`'s doc comment already documents for exactly this reason ("turns
  a broken pipe into a returned error instead of a panic"). A closed pipe now
  behaves as `export | head` already does: a clean `IO error: Broken pipe`.

`ensure_secrets` is deliberately untouched — its output is prompts and
diagnostics, not report — and the lock is released before it, so prompting is
unaffected. No change to exit codes, to what is printed, or to dependencies.

### Tests

`secretspec/tests/check_report_stream.rs`, three tests, each confirmed to fail
without the change:

- `a_passing_check_writes_its_whole_report_to_stdout`
- `a_failing_check_reports_on_stdout_but_errors_on_stderr`
- `a_reader_that_closes_the_pipe_early_does_not_panic` — `#[cfg(unix)]`, drives
  a real `sh`/`head` pipeline

The first two fail against the pre-fix tree; the third fails against a bare
`eprintln!` → `println!` swap and passes here.

### Compatibility

Filed under `Changed` rather than `Fixed`: anyone capturing via `2>&1` keeps
working, but a script capturing stderr *specifically* needs to capture stdout
instead. I won't claim that's rare — it caught me, and I had to update three
fixtures in my own downstream post-install suite.

### Open questions, happy to go either way

- The constraint-violation lines in `display_validation_errors` are arguably
  diagnostics rather than report. I kept them with the rest of the report on
  stdout, on the grounds that splitting one human-readable report across two
  streams is the same class of bug — say the word and I'll move them.
- `Secrets::check()` is public, so this makes a library write to its consumer's
  stdout. It already printed unconditionally, just to the other stream, so no
  consumer is spared it today. If you'd rather a library not print at all, the
  cleaner shape is `check` taking an `impl Write` defaulting to `io::stdout()`,
  or moving rendering into the CLI layer — happy to send either instead.
- Two adjacent pre-existing things this deliberately does not touch:
  `check --json | head` panics the same way already (`cli/mod.rs:1456`), and
  `import`'s summary is on stderr too (`secrets.rs:4313`/`:4320`/`:4328`). Glad
  to include either if you want them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Applying an active profile preserves each provider's public URI and
   storage/cache identities, so profile-aware native references continue to
   match the same provider during planning and resolution.
+- `secretspec check` writes its report to stdout instead of stderr, so
+  `secretspec check | grep DATABASE_URL` finds the report rather than silently
+  returning nothing. This also makes the existing colour handling correct:
+  `colored` decides whether to emit ANSI escapes by testing stdout, so a report
+  written to stderr could leak raw escape bytes into a redirected log, or drop
+  colour from output being read on a terminal. `check --json` and
+  `check --explain` already wrote to stdout and are unchanged. The report now
+  goes through a locked sink, so a reader that closes the pipe early
+  (`check | head`) reports a broken pipe and exits non-zero instead of
+  panicking, matching `secretspec export`. Scripts that capture the report by
+  redirecting stderr specifically need to capture stdout instead; `2>&1`
+  keeps working.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   written to stderr could leak raw escape bytes into a redirected log, or drop
   colour from output being read on a terminal. `check --json` and
   `check --explain` already wrote to stdout and are unchanged. The report now
-  goes through a locked sink, so a reader that closes the pipe early
-  (`check | head`) reports a broken pipe and exits non-zero instead of
+  goes through a sink rather than `println!`, so a reader that closes the pipe
+  early (`check | head`) reports a broken pipe and exits non-zero instead of
   panicking, matching `secretspec export`. Scripts that capture the report by
   redirecting stderr specifically need to capture stdout instead; `2>&1`
   keeps working.
--- a/secretspec/src/cli/mod.rs
+++ b/secretspec/src/cli/mod.rs
@@ -1516,7 +1516,7 @@ pub fn main() -> Result<()> {
             }
 
             let mut validated = app
-                .check(no_prompt)
+                .check(no_prompt, &mut std::io::stdout())
                 .into_diagnostic()
                 .wrap_err("Failed to check secrets")?;
             // Persist temp files so they outlive the command
--- a/secretspec/src/secrets.rs
+++ b/secretspec/src/secrets.rs
@@ -3675,30 +3675,48 @@ impl Secrets {
         self.ensure_reason_for(AuditAction::Check, None)?;
         let profile_display = self.resolve_profile_name(None);
 
-        eprintln!(
+        // The report is this command's output, so it goes to stdout, matching
+        // the `--json` and `--explain` paths above it.
+        //
+        // Written through a sink rather than `println!` for the reason
+        // `write_export` gives: a closed pipe (`check | head`) becomes a
+        // returned error instead of a panic. Locked once so the whole report is
+        // one contended acquisition rather than one per line.
+        use std::io::Write;
+        let stdout = io::stdout();
+        let mut out = stdout.lock();
+
+        writeln!(
+            out,
             "Checking secrets in {} (profile: {})...\n",
             self.config.project.name.bold(),
             profile_display.cyan()
-        );
+        )?;
 
         // Validate and display results
         // The read is audited inside `validate()`, so no bulk event here.
         match self.validate()? {
             Ok(valid) => {
-                self.display_validation_success(&valid)?;
+                self.display_validation_success(&mut out, &valid)?;
                 // All secrets present - return early without re-validating
                 Ok(valid)
             }
             Err(errors) => {
-                self.display_validation_errors(&errors)?;
+                self.display_validation_errors(&mut out, &errors)?;
+                // Release the lock before `ensure_secrets`, which prompts.
+                drop(out);
                 // Missing secrets - prompt if interactive (and not no_prompt) and re-validate
                 self.ensure_secrets(None, None, !no_prompt)
             }
         }
     }
 
-    /// Display validation success results
-    fn display_validation_success(&self, valid: &ValidatedSecrets) -> Result<()> {
+    /// Display validation success results to `out` (see `check`).
+    fn display_validation_success(
+        &self,
+        out: &mut dyn io::Write,
+        valid: &ValidatedSecrets,
+    ) -> Result<()> {
         let mut found_count = 0;
         let mut optional_count = 0;
         let default_names = valid
@@ -3712,23 +3730,37 @@ impl Secrets {
             let label = format_secret_label(name, config.description.as_deref());
             if missing_optional.contains(&name) {
                 optional_count += 1;
-                eprintln!("{} {} {}", "○".blue(), label, "(optional)".blue());
+                writeln!(out, "{} {} {}", "○".blue(), label, "(optional)".blue())?;
             } else if config.default.is_some() && default_names.contains(&name) {
                 found_count += 1;
-                eprintln!("{} {} {}", "○".yellow(), label, "(has default)".yellow());
+                writeln!(
+                    out,
+                    "{} {} {}",
+                    "○".yellow(),
+                    label,
+                    "(has default)".yellow()
+                )?;
             } else {
                 found_count += 1;
-                eprintln!("{} {}", "✓".green(), label);
+                writeln!(out, "{} {}", "✓".green(), label)?;
             }
         }
 
-        eprintln!("\n{}", Self::format_summary(found_count, 0, optional_count));
+        writeln!(
+            out,
+            "\n{}",
+            Self::format_summary(found_count, 0, optional_count)
+        )?;
 
         Ok(())
     }
 
-    /// Display validation error results
-    fn display_validation_errors(&self, errors: &ValidationErrors) -> Result<()> {
+    /// Display validation error results to `out` (see `check`).
+    fn display_validation_errors(
+        &self,
+        out: &mut dyn io::Write,
+        errors: &ValidationErrors,
+    ) -> Result<()> {
         let mut found_count = 0;
         let mut missing_count = 0;
         let mut optional_count = 0;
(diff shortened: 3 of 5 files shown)


domenkozar (commented):
Reviewed this against a local checkout and ran the tests. The stream split is the right call, the colour argument holds up, and the description saved me most of the digging. Four things, one of which answers your second open question.

### 1. Take the sink as a parameter rather than reaching for stdout inside the library

Yes to the `impl Write` shape you offered. `export` already established it: `cli/mod.rs:1466` acquires stdout in the CLI and passes `&mut out` into `Secrets::export`. `check` should look the same:

```rust
pub fn check(&self, no_prompt: bool, out: &mut dyn io::Write) -> Result<ValidatedSecrets>
```

with `cli/mod.rs:1519` handing it the stream. One caller in the tree and nothing in `secretspec-derive` touches `check`, so the churn is small, and a published library then stays out of its consumer's data channel.

One detail when you move it: pass `&mut std::io::stdout()` rather than a held `.lock()`. Holding the lock across `self.validate()` is the part of the current shape I would not keep. `validate()` fans out onto `std::thread::scope` workers whenever a check spans more than one provider group (`secrets.rs:5412`), so a global lock is held across arbitrary provider I/O running on other threads. Nothing in the tree writes to stdout from those workers today (no provider prints, none use `Stdio::inherit`), so this is latent rather than a reachable hang, but the lock buys no atomicity when nothing else writes to the stream, and an unlocked handle re-acquires per line exactly like `println!` does. `export` has the same pattern already; pre-existing, not yours to fix here.

### 2. The pipeline test earns its place, and I confirmed it is not vacuous

The report is small enough to fit in a 64KB pipe buffer, so I wanted to be sure `head -1` really does close the pipe before the writer is done rather than the whole report landing in the buffer first. Measured against a build of this branch:

* 100 runs of that exact pipeline: 99 produced `Broken pipe`, so the post-header writes really do hit a closed pipe.
* Substituting a panicking sink for yours (a stand-in for the naive `println!` swap): the test failed 10 out of 10 runs.

The timing works because the header is flushed before `validate()` runs, and `validate()` then does a provider read plus audit writes, which is the whole window `head` needs to print and exit. Buffer capacity never comes into it. Roughly 1 run in 100 will not exercise the regression path, which I am happy with; making it deterministic would mean a manifest large enough to fill the pipe buffer, and that is not worth the test's runtime.

### 3. A closed pipe still prints an error on the pipeline this change exists to enable

`check | head` now exits 1 with `Failed to check secrets: IO error: Broken pipe` on stderr. Nothing in the repo touches `SIGPIPE`, so restoring the default disposition in the CLI entry point would quiet that for `check`, `check --json`, and `export` in one place, and would cover the pre-existing `--json | head` panic you flagged without visiting each call site. It needs a unix-only `libc` dependency (already in the lock file transitively), so a follow-up PR is the right size for it. Not blocking this one.

### 4. Two small ones

* `tests/check_report_stream.rs:181`: `env` is interpolated raw while `bin` and `manifest` go through `shell_quote`. The surrounding single quotes cover it in practice, but the helper is right there and the comment above it makes the case.
* Changelog: "goes through a locked sink" is implementation detail in a user-facing entry, and it stops being accurate after point 1. The observable half, a closed pipe reporting an error instead of panicking, already carries the meaning.

On your first open question: keeping the constraint violations on stdout with the rest of the report is right, they are report lines rather than diagnostics. On the third: leave `import`'s summary alone, it is a separate decision and point 3 above covers the `--json` half.

---

🤖 Review generated with [Claude Code](https://claude.com/claude-code). Noted here because the GitHub API has no field for it: https://github.com/cli/cli/issues/13904

djbclark:
Thanks for the thorough review — all addressed in 1b1783c:

1. `check` now takes `out: &mut dyn io::Write` instead of locking stdout internally, matching `export`'s shape exactly (`cli/mod.rs:1466`). The CLI passes `&mut std::io::stdout()`, not a held lock — you're right that it bought no atomicity and would have serialized against anything the `validate()` worker threads might someday write.
2. No action needed — thanks for actually measuring it.
3. Agreed this is separate-PR sized; leaving SIGPIPE disposition alone here.
4. Fixed: `tests/check_report_stream.rs` now `shell_quote`s the env path like its neighbors, and the changelog no longer says "locked sink" now that the lock is gone.

djbclark:
The SIGPIPE follow-up from point 3 is now open as #377.

It turned out to be worth doing on its own: `check --json` doesn't just report a broken pipe, it *panics* out of `println!` (exit 101). Resetting the disposition at the entry point covers that and `export` together, and will cover `check` here once this PR lands.

domenkozar:
This one is still needed?

PR #374, opened 2026-08-17
Format-preserving single-declaration edits on `Spec`

### LINKED ISSUE #370: Format-preserving single-declaration edits on `Spec` — following up on your note in #356/#357
Opening this at your invitation on #356 and #357, and deliberately shaped around
your reasoning on #357 rather than re-proposing the same thing a third time:

> I mean that the public api will be `Spec` and `Spec::from(path)` instead of
> the internal Config.

That's the right call and I'm not asking you to reverse any of it. `Config`
should stay internal. What follows lands entirely on `Spec`.

Two things from #356 I'm explicitly **not** re-raising, because #334 settled
them: tri-state requiredness (`Secret::optional`/`required`, and
`required_setting() -> Option<bool>`) is exactly the semantics I was reaching
for, and it landed better in #334 than in my PR.

### ELI5

`Spec` can **read** a `secretspec.toml`, and can **build** a new one from
scratch. It cannot **write an existing one back out**.

So if you already have a real file on disk — with a team's comments, their key
order, their quoting — and you want code to add or remove one declaration in it,
there's no supported way to do that. You either regenerate the whole file from
the parsed model (losing every comment and reordering every profile) or you
hand-roll `toml_edit` yourself outside the library.

**The ask:** two narrow methods on `Spec` that perform the one-line edit and
leave every other byte of the file exactly as it was.

### Technically

`Spec { config, compiled, base_dir }` has no `to_toml`, no `Display`, and no
`toml_edit` anywhere in `spec.rs`. `SpecBuilder::build()` goes through
`Spec::from_config_document(self.config)` — an in-memory `Config` that is never
written anywhere. `Config.profiles` is a `HashMap`, so key order isn't preserved
even in principle. There is currently no serialization path from `Spec` back to
a file at all, exact or otherwise.

Proposed surface — no new public type, no `Config` and no `toml_edit` in any
signature:

```rust
impl Spec {
    /// Add one declaration, returning a new fully-revalidated `Spec` whose
    /// preserved text differs from the original only by that declaration.
    pub fn add_secret_to_text(&self, profile: &str, name: &str, secret: Secret) -> Result<Spec>;

    /// The inverse, same guarantee.
    pub fn remove_secret_from_text(&self, profile: &str, name: &str) -> Result<Spec>;

    /// Whether `profile` declares `name` in this spec's own source text
    /// (not only via inheritance).
    pub fn declares_secret_in_text(&self, profile: &str, name: &str) -> bool;

    /// Render as TOML, freshly formatted. Always available.
    pub fn to_toml(&self) -> String;

    /// This spec's exact backing text, when the byte-exactness guarantee still
    /// holds. `None` for a `Spec::builder()`-constructed spec.
    pub fn preserved_text(&self) -> Option<&str>;
}
```

Guarantee: `Spec::from_toml(s)?.preserved_text() == Some(s)`, and add-then-remove
of the same key returns the original bytes.

**Implementation shape that makes this safe.** These methods edit the retained
source text with `toml_edit`, then re-derive `config`/`compiled` by reparsing the
result through the same validated path every other `Spec` already goes through.
The semantic view is always *derived* from the text, never hand-mutated in
parallel with it — one synchronization point, not one per method, and no
possibility of the two disagreeing.

**Deliberately not proposed:** making `SpecBuilder`'s general edit surface
format-preserving. `replace_secret`, `profile`, `provider`, `scope` and the
`Secret`/`Profile` setters have no single canonical textual form (does replacing
a value reorder an inline table's keys, or append?), and pretending otherwise
means maintaining two hand-mutated representations in lockstep forever, for
every future schema field. `into_builder()`/`to_builder()` stays a hard
boundary: only `Spec` carries source text.

For the same reason `to_toml()` and `preserved_text()` are two methods rather
than one. A single polymorphic `to_toml()` whose exactness depends on hidden
state is a footgun for exactly the callers who need the guarantee — they'd get a
silently regenerated document. Making the exact path return `Option` forces the
question at the type level.

**Precedent.** This is the problem `cargo add` has, and I'd suggest the same
resolution. Cargo has a serde model of `Cargo.toml` *and* needs
formatting-preserving edits; the team explicitly didn't want two TOML parsers,
so `cargo-edit` performs `toml_edit` surgery kept separate from the serde
`Manifest` rather than fused into it. They still invest in it
(rust-lang/cargo#12838, "fix(add): Preserve more comments"), and epage maintains
both `toml_edit` and `cargo-add`.

**Most of this already exists here, privately.** `cli/mod.rs:618` is
`add_secret_to_manifest(source, profile, name, description)` — `toml_edit`-backed,
whose own doc comment says it "retains the user's comments, whitespace,
ordering, and any syntax that is not represented by `Config`."
`cli/mod.rs:480` is `generate_toml_with_comments(&Config)`, a complete
deterministic writer used by `init`. And `toml_edit = "0.23"` is already a
workspace dependency, just gated `cli = ["dep:toml_edit", ...]`. So this is
mostly promotion and a feature-gate change (`manifest-edit = ["dep:toml_edit"]`,
`cli = ["manifest-edit", ...]`), not new machinery. Removal is the one genuinely
new capability — and it's what makes the round-trip *provable*, since you can't
assert byte-exact undo with `add` alone.

**One wrinkle I'd rather flag than have surface at review.** `Spec::from_toml`
rejects a document with a non-empty `project.extends`, having no path to resolve
against; `Spec::try_from(path)` handles it and already stores `base_dir`. So the
internal reparse needs an `extends`-aware variant seeded with `self.base_dir`.
Solvable with an internal helper shared with `try_from` — no new public API —
but it's real work, not a one-liner. Relatedly, the retained text must be the
**root file only**: `Config::try_from(path)` merges parents, so writing a merged
config back into the root document would silently inline inherited declarations
into the child file.

### Use cases

**1. Config-automation bots opening PRs.** Any tool that proposes a
`secretspec.toml` change as a pull request — service scaffolding, a CI job
declaring a newly-required secret when a workflow changes, a Renovate-style bot
— needs that diff to be a clean one-line addition. A `Config` round-trip can't
produce that diff by construction: it reorders every profile, strips every
comment, and collides with every other open PR touching the file. Whether the
diff is one line or a full-file rewrite decides whether the change is reviewable
at all.

**2. Editor and pre-commit tooling.** A "declare this env var" quick-fix, or a
pre-commit hook adding a missing declaration, shouldn't reformat a file a human
maintains by hand in order to add one line to it.

**3. Provable undo (my case, briefly).** I maintain a downstream fork running
`secretspec` as a privilege-separated broker, where `add` has a runtime inverse
and the guarantee is that add → undeclare restores the manifest byte for byte,
not merely to something semantically equivalent. That's testable precisely
because `toml_edit` doesn't touch what it wasn't asked to touch.

### What I'm offering

The `toml_edit` surgery exists and is tested in my fork as an extension of your
`cli/mod.rs:618`, plus the removal counterpart. The tests assert the round-trip
directly, including against a document mixing the shapes most likely to trigger
reserialization — a `required` table for a presence group, a full
`[profiles.x.NAME]` table, a quoted dotted key, and inline tables in one file —
across all three requiredness states. Plus negative cases: the "is it declared"
predicate parsed rather than substring-matched (so it isn't fooled by a comment
or by the name appearing inside another secret's description), an unparseable
manifest reported as an error rather than `false`, and removal of an undeclared
name an error rather than a silent no-op.

Porting that onto the `Spec` shape above is more work than exposing my module
directly, but it's the right amount of work given where you've said the public
surface is going. Happy to open the PR against this shape, or a different one
you'd prefer — I mainly wanted the use case on record first, as you asked.

djbclark:
Closes #370.

Opening this against the shape you described on #357 — everything lands on `Spec`, `Config` stays internal, and no `toml_edit` or `Config` appears in any public signature.

### What it does

`Spec` can read a `secretspec.toml` and can build a new one, but cannot write an existing one back out. So editing one declaration in a real file means either regenerating the whole document from the parsed model — losing every comment and reordering every profile, since `Config.profiles` is a `HashMap` — or hand-rolling `toml_edit` outside the library.

Five methods, the surface proposed in the issue:

```rust
impl Spec {
    pub fn add_secret_to_text(&self, profile: &str, name: &str, secret: Secret) -> Result<Spec>;
    pub fn remove_secret_from_text(&self, profile: &str, name: &str) -> Result<Spec>;
    pub fn declares_secret_in_text(&self, profile: &str, name: &str) -> bool;
    pub fn preserved_text(&self) -> Option<&str>;
    pub fn to_toml(&self) -> Result<String>;
}
```

`Spec::from_toml(s)?.preserved_text() == Some(s)`, and add-then-remove of the same key returns the original bytes.

### How it stays safe

The edits are `toml_edit` surgery on the retained text; `config`/`compiled` are then re-derived by **reparsing the result through the same validated path every other `Spec` already goes through**. The semantic view is always derived from the text, never hand-mutated alongside it — one synchronization point instead of one per method, so the two cannot disagree, and an edit that would not validate fails at the edit rather than at some later load.

Per the issue, `SpecBuilder`'s general edit surface is **not** made format-preserving. `into_builder()`/`to_builder()` stays a hard boundary: only `Spec` carries source text.

### The two wrinkles I flagged, and how they're handled

**1. `extends`.** `Spec::from_toml` rejects a non-empty `project.extends`, having nowhere to resolve paths from, so reparsing edited text through it would fail on every inheriting project. `Config::from_text_in` is the `extends`-aware variant, seeded with the `base_dir` `Spec` already records. `ConfigGraphLoader` gained an in-memory entry point, and its `extends` walk is now shared with the path-based one rather than duplicated.

**2. Root file only.** `Config::try_from` folds parents into the child, so retaining the *merged* text and writing it back would silently inline every inherited declaration into a file that had merely referenced them. A test asserts the parent's declaration does not appear in the child's text while `secrets()` still sees it.

### Module and feature

The `toml_edit` surgery moves out of `cli` into a `manifest_edit` module behind a new `manifest-edit` feature, which `cli` enables — nothing changes for existing users, but an embedder can take the editing surface without `clap` and `inquire`.

`toml_edit` gains its `serde` feature so a whole `Secret` renders through the same representation the parser reads, keeping written and accepted keys from drifting as the schema grows. The value serializer refuses nested tables — which `ref`, `refs`, `extract`, `generate` and a presence group's `required` all produce — so declarations serialize via `to_document` and are flattened to an inline table. `Cargo.lock` gains only `serde_core` and `serde_spanned`, both already in the tree via `toml`; no new crates.

**Requiredness is untouched.** As I said on the issue, #334 settled it — `add_secret_to_manifest` keeps its existing description-only signature, and anything richer goes through the `Secret` that `add_secret_to_text` already takes.

### Two deviations from the issue's sketch, both deliberate

- `to_toml` returns `Result<String>`, not `String`. TOML serialization is fallible and I would rather surface that than unwrap inside the library.
- The three editing methods are `#[cfg(feature = "manifest-edit")]`, since they need `toml_edit`. `preserved_text` and `to_toml` are unconditional.

Happy to change either, or the naming, if you'd prefer something else.

### Tests

New tests cover the byte-exact round-trip; that comments, declaration order, and a full `[profiles.x.NAME]` table all survive an edit; that the returned `Spec` is revalidated rather than merely rewritten, and that the spec it derived from is untouched; that an invalid edit fails at the edit; that a declaration carrying `providers`, `as_path` and a nested `ref` round-trips; duplicate-add and absent-remove errors; `declares_secret_in_text` parsed rather than substring-matched, so a name inside a comment or another secret's description does not fool it; and a builder-built spec reporting no text rather than inventing one.

Four more cover inheritance against real files on disk: root-only retention, editing a spec that `extends` another, an inherited declaration not being editable in the child, and the round-trip holding through inheritance.

Full suite green on this branch. The 21 `provider::sops::*` failures in my environment are all `The 'sops' CLI is not installed` and occur identically on an unmodified `dfa4b10`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Spec` can perform format-preserving single-declaration edits on the document
+  it was loaded from. `add_secret_to_text` and `remove_secret_from_text` return
+  a new fully revalidated `Spec` whose text differs only by that declaration,
+  leaving comments, key order, quoting, and syntax the model does not represent
+  byte for byte intact; `declares_secret_in_text` reports whether a name is
+  declared in this document rather than inherited; `preserved_text` exposes the
+  exact backing text, and `to_toml` renders freshly formatted TOML for any spec,
+  including one built with `Spec::builder`, which has no backing text. Adding a
+  declaration and then removing it restores the original bytes, so tooling that
+  proposes a manifest change as a reviewable one-line diff, or that compares
+  manifests byte for byte, can rely on it. A spec that inherits through
+  `project.extends` keeps its own root document as the preserved text and is
+  revalidated against its parents on every edit, so inherited declarations are
+  never inlined into the child file. The `toml_edit` surgery behind this moved
+  out of `cli` into a `manifest_edit` module behind a new `manifest-edit`
+  feature, so an embedder can take the editing surface without `clap` and
+  `inquire`; `cli` enables it, so nothing changes for existing users.
+
 - **Azure App Configuration provider** (`aac://`, 0.20+): select direct
   values and Azure Key Vault references by label, prefix, and tags, with Entra
   ID or connection-string authentication and guarded writes, deletion, and
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6112,6 +6112,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "84c8b9f757e028cee9fa244aea147aab2a9ec09d5325a9b01e0a49730c2b5269"
 dependencies = [
  "indexmap 2.13.0",
+ "serde_core",
+ "serde_spanned",
  "toml_datetime 0.7.5+spec-1.1.0",
  "toml_parser",
  "toml_writer",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,10 @@ keepass = { version = "0.13.17", features = ["save_kdbx4"] }
 keeper-secrets-manager-core = { version = "17.3.0", default-features = false, features = ["blocking"] }
 serde = { version = "1.0", features = ["derive"] }
 toml = "0.9"
-toml_edit = "0.23"
+# `serde` powers `manifest_edit`'s full-declaration writer, which renders a
+# `Secret` through the same representation the parser reads so the written and
+# accepted keys cannot drift apart as the schema grows.
+toml_edit = { version = "0.23", features = ["serde"] }
 thiserror = "2.0"
 etcetera = "0.11"
 colored = "3.0"
--- a/secretspec/Cargo.toml
+++ b/secretspec/Cargo.toml
@@ -60,7 +60,10 @@ age = { workspace = true, optional = true }
 
 [features]
 default = ["cli", "keyring", "kdbx", "keeper", "gcsm", "awssm", "awsps", "vault", "openbao", "bws", "akv", "aac", "infisical", "bw", "age", "scaleway", "sops"]
-cli = ["dep:toml_edit", "dep:clap_complete", "dep:clap_complete_nushell", "dep:is_executable"]
+cli = ["manifest-edit", "dep:clap_complete", "dep:clap_complete_nushell", "dep:is_executable"]
+# Format-preserving declaration edits, without the interactive CLI. Split out
+# so an embedder can take the editing surface without taking `clap`/`inquire`.
+manifest-edit = ["dep:toml_edit"]
 keyring = ["dep:keyring", "dep:whoami"]
 kdbx = ["dep:keepass"]
 keeper = ["dep:keeper-secrets-manager-core"]
--- a/secretspec/src/cli/mod.rs
+++ b/secretspec/src/cli/mod.rs
@@ -1,4 +1,5 @@
 use crate::config::{Config, GlobalConfig, GlobalDefaults, Profile as ConfigProfile, Project};
+use crate::manifest_edit::{add_secret_to_manifest, validate_add_secret_name};
 use crate::provider::{Provider, providers, spec_names_known_provider};
 use crate::{CallerContext, ExportFormat, Secrets};
 use clap::{Parser, Subcommand, ValueEnum, ValueHint};
@@ -566,25 +567,6 @@ fn generate_toml_with_comments(config: &Config) -> crate::Result<String> {
     Ok(doc.to_string())
 }
 
-/// Rejects names that cannot occupy a flattened secret key in [`ConfigProfile`].
-fn validate_add_secret_name(name: &str) -> Result<()> {
-    if !crate::config::is_valid_identifier(name) {
-        return Err(miette!(
-            "Invalid secret name '{}': must be a valid identifier (alphanumeric and underscores, not starting with a number)",
-            name
-        ));
-    }
-    // `Profile` reserves this key for its defaults table before flattening all
-    // remaining keys into secret declarations. Without an explicit check, the
-    // edit would be valid TOML but would not actually declare a secret.
-    if name == "defaults" {
-        return Err(miette!(
-            "Secret name 'defaults' is reserved for profile defaults"
-        ));
-    }
-    Ok(())
-}
-
 /// Ensures `add` will create a new effective declaration in an existing profile.
 fn validate_add_target(app: &Secrets, profile: &str, name: &str) -> Result<()> {
     validate_add_secret_name(name)?;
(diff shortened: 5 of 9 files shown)


djbclark:
I've put this API through a real consumer before review rather than after, so
here's a dogfooding report from a downstream fork that maintains a
root-privileged manifest editor.

**Context:** the consumer is a privilege-separated broker that owns a
root-owned `secretspec.toml` in a protected vault. Its `add`/`undeclare` verbs
were calling the old `manifest_edit` helpers directly. I ported them to the
`Spec` surface this PR proposes.

**What worked without adjustment**

- `Secret::new` / `Secret::required` / `Secret::optional` turn out to express
  exactly the tri-state I had been passing around as `Option<bool>` — leave
  requiredness to the profile default, or pin it either way. I had assumed
  adopting the `Secret`-shaped API would cost me that distinction, and it
  doesn't. Worth noting for anyone else weighing the same migration.
- The byte-exactness holds under the test I care about most: add a declaration
  and remove it again, and the document is restored byte for byte. My undo path
  compares manifests as bytes, so a round trip that merely preserved *meaning*
  would be useless to me. This is now asserted downstream.
- Revalidating the edited document as a whole is a real improvement for a
  privileged writer: a declaration that would not load is refused before
  anything reaches the vault, instead of surfacing at the next `check`.

**`Spec::from_toml` is load-bearing for a privileged editor, somewhat by accident**

It's the constructor I want in a root process specifically because it never
touches the filesystem and refuses `project.extends` — so an edit cannot be
induced into reading whatever a parent path points at. Right now that property
falls out of the extends refusal rather than being stated as a guarantee. It
might be worth saying so in the doc comment: "does not read the filesystem" is
the reason a caller in a privileged context picks this over
`TryFrom<&Path>`, and it's currently something you have to derive by reading
`reparse`.

**One call site I deliberately did not port**

A guard that asks whether a *different* document — a tracked template the
editor doesn't hold as a `Spec` — declares a given name, and which must fail
closed: an unparseable template is not evidence that a name is absent from it,
and the guard exists to protect exactly the names it might have failed to read.

`Spec::declares_secret_in_text` answers `bool`, so an unrepresentable document
reads as "not declared". In practice that branch is near-unreachable, since
constructing the `Spec` already parsed the source — this is not a bug report.
But the predicate is about the spec's *own* text, and my question is about
someone else's; routing it through a `Spec` would also mean full validation and
an `extends` refusal on a document that is allowed to inherit.

So that one stays on `manifest_edit::declares_secret`, which takes text and
returns `Result`. This works only because the PR promotes `manifest_edit` to a
public module behind its own feature. Flagging it in case that surface is ever
considered for narrowing: the `Spec` methods and the text functions serve
genuinely different callers, and at least one real consumer needs both.

Happy to share the downstream diff if it's useful.

domenkozar:
Superseded by #389, rebuilt from current main around the existing SpecBuilder API and toml_edit.

PR #384, opened 2026-08-19
fix(onepassword): batch shared-item field reads via op inject

ifiokjr:
## Why

`monosecret run` / `msload` were taking ~37s to load secrets from 1Password. The installed binary resolved each secret with its own `op read` plus an `op vault list` preflight per secret — 32 `op` process spawns for a profile whose secrets are all fields of **one** shared item (`Dotfiles` in vault `Development`, organized by sections). That serial, per-secret loading is what made it slow.

The source already batches field references into a single `op inject` call, but the commit that added that batching also dropped the legacy `op+token://` scheme, so existing configs couldn't reach the fast path.

## Implementation

In `crates/monosecret/src/provider/onepassword.rs`:

- Restore the legacy `op+token://<account>/<basePath>` scheme and parse its path into a `reference_base_path` on `OnePasswordConfig`. A secret at `path = ["ai"]` now resolves to item `Dotfiles`, section `ai` (via `apply_reference_base_path` in `operation_coordinates`).
- Route these through the existing batched `get_many` → `op inject` path, so all field references are fetched in **one** call plus a single auth preflight.
- Keep `onepassword+token://` behavior unchanged (token still supplied via provider credential or `OP_SERVICE_ACCOUNT_TOKEN`); the token-in-URI rejection now also covers `op+token://`.
- Add regression tests for `op+token://` parsing, token-in-URI rejection, and base-path folding.

Result: **~2–5s vs ~37s**, **2 `op` invocations vs 32**, all secrets resolved.

_Created on behalf of Ifiok Jr. ([@ifiokjr](https://github.com/ifiokjr)) by codex using DeepSeek (deepseek-v4-flash:cloud) at default thinking._


ifiokjr:
Closing: this PR was created against the wrong repository by automation. The intended target is ifiokjr/monosecret.

PR #387, opened 2026-08-19
feat(provider): add Kubernetes

### LINKED ISSUE #316: Support kubernetes secrets as provider
Add support for using kubernetes [secrets](https://kubernetes.io/docs/concepts/configuration/secret/) as a provider

giancarlopernudisegura:
## Summary

- add Kubernetes provider support with refs, writes (including refs)
- can use either Kubernetes [ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) or [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/)
- reference Kubernetes object in specific namespace or omit for default cluster namespace
- document 0.20+ configuration

## Scenario coverage

- Secret value is stored in Kubernetes because it was generated by a helm provider
- Secret value is stored in Kubernetes and gets updated via cronjob so only accurate real-time value exists there
- User may lack permission to write to the Kubernetes ConfigMap or Secret
- User may want to reference a specific key in the Kubernetes ConfigMap/Secret via ref

## Validation

- `cargo test --all`
- `devenv shell -- prek run -a`
- Manually tested that getting and setting values worked for both ConfigMaps and Secrets, with or without explicit namespace, and with or without refs using an [AKS](https://learn.microsoft.com/en-us/azure/aks/) cluster.

Fixes #316

### DIFF
--- a/docs/src/content/docs/reference/providers.md
+++ b/docs/src/content/docs/reference/providers.md
@@ -671,7 +671,7 @@ k8s+secret://db-credentials@db-postgres
 **Prerequisites**: A Kubernetes configuration in `$KUBECONFIG` or
 `$HOME/.kube/config`; build with `--features kubernetes` (0.20+)
 **Authentication**: Configured in Kubernetes configuration
-**Storage**: `secretspec-{project}-{profile}-{key}` key under `.data` in the
+**Storage**: `secretspec--{project}--{profile}--{key}` key under `.data` in the
 Kubernetes object
 
 ## Provider Selection
--- a/docs/src/content/docs/reference/providers.md
+++ b/docs/src/content/docs/reference/providers.md
@@ -651,6 +651,29 @@ per project/profile. A single-file provider supports `ref = { item = "..." }`
 as a root key (or a key in `[DEFAULT]` for INI); extra coordinates and refs
 through templated paths are rejected.
 
+
+## Kubernetes Provider (0.20+)
+
+:::caution[Version compatibility]
+The `kubernetes` provider is added in SecretSpec 0.20.
+:::
+
+**URI**: `k8s+KIND://NAME[@NAMESPACE]` - Stores secrets in a Kubernetes
+ConfigMap or Secret
+
+```text
+k8s+configmap://db-config@db-postgres
+k8s+configmap://db-config
+k8s+secret://db-credentials@db-postgres
+```
+
+**Features**: Read/write Kubernetes ConfigMaps and Secrets
+**Prerequisites**: A Kubernetes configuration in `$KUBECONFIG` or
+`$HOME/.kube/config`; build with `--features kubernetes` (0.20+)
+**Authentication**: Configured in Kubernetes configuration
+**Storage**: `secretspec-{project}-{profile}-{key}` key under `.data` in the
+Kubernetes object
+
 ## Provider Selection
 
 ### Command Line
@@ -711,3 +734,4 @@ $ export SECRETSPEC_PROVIDER="dotenv:///config/.env"
 | Infisical | ✅ Infisical-managed | Cloud (Infisical) or self-hosted | ✅ Yes |
 | age (0.17+) | ✅ age encryption | Local filesystem | ❌ No |
 | SOPS (0.17+) | ✅ Configured SOPS encryption | Local filesystem | Depends on configured key service |
+| Kubernetes (0.20+) | ❌ ConfigMap ✅ Secret if configured | Kubernetes server | ✅ Yes |
--- a/secretspec/src/provider/kubernetes.rs
+++ b/secretspec/src/provider/kubernetes.rs
@@ -152,6 +152,15 @@ impl KubernetesProvider {
         }
     }
 
+    pub fn build_uri(kind: &KubernetesKind, name: &String, namespace: &Option<String>) -> String {
+        let mut uri = format!("k8s+{}://{}", kind, name);
+        if let Some(namespace) = namespace {
+            uri.push('@');
+            uri.push_str(namespace);
+        }
+        uri
+    }
+
     async fn client(&self) -> Result<&Client> {
         if let Some(client) = self.client.get() {
             return Ok(client);
@@ -384,12 +393,7 @@ impl Provider for KubernetesProvider {
     }
 
     fn uri(&self) -> String {
-        let mut uri = format!("k8s+{}://{}", self.config.kind, self.config.name);
-        if let Some(namespace) = &self.config.namespace {
-            uri.push('@');
-            uri.push_str(&namespace);
-        }
-        uri
+        Self::build_uri(&self.config.kind, &self.config.name, &self.config.namespace)
     }
 
     fn convention_address(
--- a/secretspec/src/provider/kubernetes.rs
+++ b/secretspec/src/provider/kubernetes.rs
@@ -166,35 +166,51 @@ impl KubernetesProvider {
         }
     }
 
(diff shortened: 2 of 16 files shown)


secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+        Ok(_) => std::thread::scope(|scope| {
+            let worker = scope.spawn(move || runtime().block_on(future));
+            match worker.join() {
+                Ok(output) => output,
+                Err(panic) => std::panic::resume_unwind(panic),
+            }
+        }),
+        Err(_) => runtime().block_on(future),
+    }
+}
```
giancarlopernudisegura, line 54: Just to be upfront, this snippet is essentially a copy/paste from [vault_common.rs](https://github.com/cachix/secretspec/blob/main/secretspec/src/provider/vault_common.rs). I don't know if it should be extracted to some common function or if just mentioning in a comment that it's essentially a duplicate is fine. Initially I just used `super::block_on` but that caused panics when trying to set a secret value.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+        let client = self.client().await?;
+        let namespace = match &self.config.namespace {
+            Some(ns) => ns.as_str(),
+            None => client.default_namespace(),
+        };
+        let spec = SelfSubjectAccessReviewSpec {
+            resource_attributes: Some(ResourceAttributes {
+                namespace: Some(namespace.into()),
+                verb: Some("patch".to_string()),
+                resource: Some(self.config.kind.to_string()),
```
domenkozar: **[P1] Use plural resource names in access reviews**

Kubernetes RBAC checks use plural API resource names. `Display` yields `secret` or `configmap` here, so ordinary roles granting `resources: ["secrets"]` or `resources: ["configmaps"]` will not match this `SelfSubjectAccessReview`, and authorized writes will be rejected. Map each kind to its plural API resource name for this field.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+    config: KubernetesConfig,
+    client: OnceLock<Client>,
+}
+
+crate::register_provider! {
+    struct: KubernetesProvider,
+    config: KubernetesConfig,
+    name: "kubernetes",
+    description: "Kubernetes",
+    schemes: ["k8s+configmap", "k8s+secret"],
```
domenkozar, line 142: **[P1] Keep the bare name out of provider selection**

The global provider picker persists `info.name`, so this registration offers `kubernetes` even though the registry only accepts the two configured URI schemes. Selecting it produces a provider value that later construction—and `config global init --provider kubernetes` itself—rejects. Please either exclude this provider from bare-name selection or make the flow collect a concrete Kubernetes URI.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+    examples: ["k8s+secret://db-config", "k8s+configmap://db-config@default"],
+    deletes: true,
```
domenkozar, line 144: **[P2] Stop advertising deletion until it is implemented**

This advertises static delete support, but `KubernetesProvider` implements neither `delete` nor `supports_delete`, so it inherits the unsupported-operation behavior. That admits cache invalidation and `import --delete-source` planning only for instance deletion to fail. Remove this flag until deletion is implemented, or implement the capability consistently.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    fn set(&self, addr: Address<'_>, value: &SecretString) -> Result<()> {
+        self.check_writable(addr)?;
+        let coords = self.resolve_coords(addr)?;
+        block_on(self.set_secret_async(&coords.item, value))
+    }
+
+    fn check_writable(&self, _addr: Address<'_>) -> Result<()> {
+        let can_i_patch = block_on(self.can_i_patch())?;
```
domenkozar: **[P2] Resolve the address during write preflight**

`_addr` is ignored, so `check_writable` can succeed for a native address containing an unsupported coordinate such as `field`; `set` then reaches `resolve_coords` and rejects it only after the CLI may have prompted for a value. Resolve the address here before the permission review so preflight rejects everything `set` will reject.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+                    c
+                )));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn format_secret_name(project: &str, profile: &str, key: &str) -> Result<String> {
+        let secret_name = format!("secretspec-{}-{}-{}", project, profile, key);
```
domenkozar: **[P2] Encode convention components without collisions**

Hyphen concatenation is not injective: `(project="a-b", profile="c", key="KEY")` and `(project="a", profile="b-c", key="KEY")` both become `secretspec-a-b-c-KEY`. Projects sharing the Kubernetes object can therefore read or overwrite one another. Use an injective component encoding or a reserved delimiter so the profile-aware path remains isolated.

secretspec/src/provider/kubernetes.rs
```diff
@@ -0,0 +1,427 @@
...
+pub struct KubernetesProvider {
+    config: KubernetesConfig,
+    client: OnceLock<Client>,
+}
+
+crate::register_provider! {
+    struct: KubernetesProvider,
+    config: KubernetesConfig,
+    name: "kubernetes",
+    description: "Kubernetes",
```
domenkozar: **[P2] Add the release version to the CLI description**

This description feeds the `config init` selector, so it currently renders `kubernetes: Kubernetes` without the release marker. Because this provider targets the unreleased 0.20 release, label the registration description `Kubernetes (0.20+)`, matching the version required at every selector entry.

domenkozar (commented):
Six actionable inline findings on the Kubernetes provider implementation.

secretspec/src/provider/kubernetes.rs
cccs-gpernud: Implemented deletion in 75d67f684eb67bbdf6b7fd251585026081a1bf60

docs/src/content/docs/reference/providers.md
```diff
@@ -651,6 +651,29 @@ per project/profile. A single-file provider supports `ref = { item = "..." }`
...
+k8s+configmap://db-config@db-postgres
+k8s+configmap://db-config
+k8s+secret://db-credentials@db-postgres
+```
+
+**Features**: Read/write Kubernetes ConfigMaps and Secrets
+**Prerequisites**: A Kubernetes configuration in `$KUBECONFIG` or
+`$HOME/.kube/config`; build with `--features kubernetes` (0.20+)
+**Authentication**: Configured in Kubernetes configuration
+**Storage**: `secretspec-{project}-{profile}-{key}` key under `.data` in the
```
domenkozar: Kubernetes Secret and ConfigMap `.data` fields are flat key/value maps rather than folder hierarchies, and `/` is not valid in a data key. The implementation reserves `--` to separate the SecretSpec coordinates, so this should match it: `secretspec--{project}--{profile}--{key}`.

secretspec/src/provider/kubernetes.rs
giancarlopernudisegura: Addressed in 706063f999b3e296a9073a6a1e08d67c3dba8b6c

domenkozar:
Merged to master with some fixes

Issue #383, opened 2026-08-19
Add musl targets: the Node SDK and the CLI cannot run on Alpine


moltar:
## Summary

SecretSpec has no musl target. The Node SDK cannot load on an Alpine image, and the CLI binary
cannot run there either. Alpine is a common base for Node services, so this blocks SecretSpec from
containers that use it.

## The Node SDK

The npm package declares only glibc addons:

```
secretspec-linux-x64-gnu
secretspec-linux-arm64-gnu
secretspec-darwin-arm64
secretspec-win32-x64-msvc
```

napi platform packages carry a libc constraint, so npm correctly skips them on musl. The install
looks successful and the failure appears at require time:

```
$ docker run --rm node:24-alpine sh -c \
    'cd /tmp && npm init -y && npm i secretspec@0.19.1 && node -e "require(\"secretspec\")"'

added 1 package, and audited 2 packages in 986ms

failed to load the secretspec native addon for linux-arm64
(tried ./secretspec.node and the 'secretspec-linux-arm64-gnu' package).
Underlying error: Cannot find module 'secretspec-linux-arm64-gnu'
```

The main package installs, so a lockfile and a CI install both pass. The error only appears when the
service starts.

## The CLI

Every CLI asset in the v0.19.1 release is `unknown-linux-gnu`. There is no musl asset. The release
does ship `secretspec-go-static-x86_64-linux-musl.tar.gz` for the Go binding, so a musl toolchain is
already in the release pipeline.

## The glibc builds also need a recent glibc

This surprised me, and it narrows the glibc workaround. The `aarch64-unknown-linux-gnu` binary does
not run on Debian bookworm:

```
$ docker run --rm node:24-bookworm-slim ...
secretspec: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by secretspec)
secretspec: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by secretspec)
```

Bookworm has glibc 2.36. The binary works on Debian trixie, which has glibc 2.41. So "use a glibc
image" means a recent one, and images built on bookworm or on Ubuntu 22.04 fail.

## What I am asking for

1. A musl target for the npm platform packages, so `require("secretspec")` works on `node:alpine`.
2. A musl target for the CLI release assets.
3. If it is cheap, build the glibc assets against an older glibc so they run on current long term
   support distributions.

## Why it matters

[...]

PR #388, opened 2026-08-20
feat(node): publish musl builds of the native addon

moltar:
`require("secretspec")` fails on `node:alpine`. The npm package declares four platform packages and all four are glibc. napi platform packages carry a `libc` constraint, so npm skips them on musl, the install reports success, and the failure only appears when the service starts. Alpine is a common base for Node services, so this blocks the SDK from those containers. #383 has the reproduction.

This adds `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` to `napi.targets` and builds them in the pinned `musllinux_1_2` containers. Those images are Alpine based and match their runner architecture, so both legs build and test natively against musl with no cross toolchain. `dotnet-package.yml` already builds `secretspec-ffi` for the same two triples the same way, and both targets are already in `rust-toolchain.toml`, so the dependency graph needed nothing. There are no Rust source changes here.

The musl legs sit in their own `musl` job rather than in the `addon` matrix, because GitHub Actions runs JavaScript actions inside an Alpine container only on x64 runners. The job stays on the host and drives the build through `docker run`, which keeps `actions/checkout` outside the container, again matching `dotnet-package.yml`. The `addon` job is unchanged apart from its header comment. `publish` waits on both.

The loader in `index.js` reads the Node diagnostic report to tell glibc from musl and picks the matching platform package. After the build, the job loads the addon on `node:24-alpine` as a separate step, because `musllinux_1_2` and `node:alpine` are different Alpine releases and the second is what a Node service actually runs.

Two things need your attention.

The two musl package names are new on npm, and npm has no pending-publisher mechanism, so each needs a manual first publish before a Trusted Publisher can attach to it. `napi pre-publish` publishes every platform package in one step, so until that bootstrap happens the publish job fails and takes the main `secretspec` package with it. I wrote the steps into RELEASE.md next to the other one-time setup notes, but only a maintainer can run them.

I left `optionalDependencies` in `package.json` alone. `napi pre-publish` derives that field from `napi.targets` at publish time, which is why the committed values still read 0.12.2 while 0.19.1 shipped with 0.19.1 there. Adding the musl names by hand breaks `npm ci` on every leg, because the lockfile cannot resolve packages that do not exist yet.

Items 2 and 3 of the issue, the musl CLI assets and the CLI glibc floor, are not here. Both live in cargo-dist config, where `install-updater` is a global setting and dist refuses a target whose updater axoupdater does not publish. That looks like the constraint `aarch64-pc-windows-msvc` already hit, so it deserves its own PR.

### DIFF
--- a/.github/workflows/node-addon.yml
+++ b/.github/workflows/node-addon.yml
@@ -6,12 +6,18 @@ name: "Node addon"
 # optionalDependencies) using @napi-rs/cli, authenticated via npm Trusted
 # Publishing (OIDC) -- no NPM_TOKEN stored in CI.
 #
-# The Linux builds run inside manylinux_2_28 containers (AlmaLinux 8, glibc
-# 2.28), so the published addon loads on any distro with glibc >= 2.28
+# The glibc Linux builds run inside manylinux_2_28 containers (AlmaLinux 8,
+# glibc 2.28), so the published addon loads on any distro with glibc >= 2.28
 # (RHEL 8/9, Amazon Linux 2023, Debian 10+, Ubuntu 18.10+). Building on the bare
 # runner instead links the runner's glibc (2.39 on ubuntu-24.04), which broke
 # RHEL 9-family distros (issue #136). A post-build step fails the job if the
 # addon's glibc floor or NEEDED libraries regress.
+#
+# The musl Linux builds run inside musllinux_1_2 containers, which are Alpine
+# based, so they build and test natively against musl with no cross toolchain.
+# npm keeps the two libc variants apart through the `libc` field in each
+# platform package. The alpine-smoke job then loads the addon on node:alpine,
+# the image the SDK must support (issue #383).
 
 on:
   workflow_call:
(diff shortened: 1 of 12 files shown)


domenkozar:
@moltar seems like CI fails

.github/workflows/node-addon.yml
```diff
@@ -58,12 +64,24 @@ jobs:
...
+          - target: linux-x64-musl
+            platform: linux-x64-musl
+            libc: musl
+            runner: ubuntu-latest
+            container: quay.io/pypa/musllinux_1_2_x86_64:2026.08.05-1@sha256:c7b2187aa4d095a8da73ea4db96acefd46701a13439a9cc4546f5d61a4b5bba1
+          - target: linux-arm64-musl
+            platform: linux-arm64-musl
+            libc: musl
+            runner: ubuntu-24.04-arm
+            container: quay.io/pypa/musllinux_1_2_aarch64:2026.08.05-1@sha256:668c455aeddf5e363bd6fd801f10b369a634a9f01974876cee8c8c518b44ef10
```
domenkozar: **[P1] Avoid Alpine job containers on ARM runners**

For the `linux-arm64-musl` matrix entry, GitHub Actions cannot execute JavaScript actions such as `actions/checkout` inside an Alpine container on an ARM64 runner; the job fails before checkout with “JavaScript Actions in Alpine containers are only supported on x64 Linux runners.” Consequently the ARM64 artifact is never built, and the dependent smoke and publish jobs cannot run. Build through an explicit container invocation from the host, or otherwise keep JavaScript actions outside this ARM Alpine job container.

moltar: Good catch, and thank you. Fixed in the follow-up commit.

The musl legs now live in their own `musl` job that runs on the host runner and
drives the build through `docker run`, so `actions/checkout` and every other
JavaScript action stays outside the Alpine container. That is the same shape
`dotnet-package.yml` already uses for its musl assets. The `addon` job is back to
exactly what it was before this PR, apart from its header comment.

Two consequences worth noting. The separate `alpine-smoke` job is gone: the new
job already builds and tests inside Alpine, so it just runs `node --test` a
second time against `node:24-alpine` as its own step, and no artifact round trip
is needed. `publish` now waits on `[addon, musl]`.

I re-ran every step of the new job verbatim on arm64 locally, substituting the
Actions expressions: the build and the SDK tests inside `musllinux_1_2`, the
portability check, and the `node:24-alpine` load. All pass.

moltar:
By the way, in full transparency I've totally wrote it with Claude. I did review and some stuff I'm familiar with like CI config. But I'm no Rust expert. Just wanted to save some time. If you don't like the solution and want to throw this out I won't be offended. 😀

PR #391, opened 2026-08-20
fix(run): forward signals to child process

### LINKED ISSUE #382: secretspec run does not forward signals, so a container app never receives SIGTERM
## Summary

`secretspec run` spawns the command as a child and never forwards signals to it. In a container,
SecretSpec becomes PID 1 and the wrapped application never receives SIGTERM. The application cannot
run its shutdown path, and the container is killed after the grace period.

I measured this in Docker. Results are below.

## Why it happens

`secretspec/src/secrets.rs:6061-6082` spawns the child and waits for it:

```rust
let child = cmd.spawn();
...
let status = child?.wait()?;
Ok(status.code().unwrap_or(1))
```

There is no `exec()` anywhere in the repository, so SecretSpec stays the parent. The crate also
installs no signal handlers. It has no `signal-hook`, no `ctrlc` and no `libc`, and `tokio` is built
without the `signal` feature. So nothing passes SIGTERM to the child.

## Measurements

Node 24 on Debian trixie, arm64, SecretSpec 0.19.1. The application logs a line in its SIGTERM
handler and then exits 0.

| Setup | `docker stop` took | Exit code | Did the app get SIGTERM |
| --- | --- | --- | --- |
| `secretspec run -- node app.js` | 10s, the full grace period | 137 | No |
| The same, with `docker run --init` | 1s | 143 | No |
| `node app.js` on its own | 0s | 0 | Yes |

In the first case SecretSpec is PID 1. The kernel removes default signal dispositions for PID 1, so
SIGTERM is discarded and Docker sends SIGKILL after the grace period.

In the second case an init process is PID 1 and forwards SIGTERM to SecretSpec. SecretSpec has no
handler, so it takes the default disposition and dies at once. The application is killed without
ever seeing the signal. Shutdown is quick, but it is not graceful.

This matters for AWS ECS, where `initProcessEnabled` gives you the second case.

## Reproduction

```js
// app.js
process.on('SIGTERM', () => { console.log('APP: got SIGTERM'); process.exit(0); });
console.log('APP: started, pid=' + process.pid);
setInterval(() => {}, 1000);
```

```toml
# secretspec.toml
[project]
name = "pid1demo"
revision = "1.0"

[providers]
e = "env://"

[profiles.default.defaults]
providers = ["e"]
required = false

[profiles.default]
DUMMY = { description = "unused" }
```

```dockerfile
FROM node:24-trixie-slim
# install the secretspec binary here
WORKDIR /app
COPY app.js secretspec.toml ./
ENTRYPOINT ["secretspec","run","--","node","app.js"]
```

Then run the container, run `docker stop`, and read the log. The application prints its start line
with a pid above 1, and it never prints the SIGTERM line.

A terminal hides this. Ctrl-C sends SIGINT to the whole foreground process group, so the child gets
it straight from the terminal driver. Only `docker stop` shows the gap.

## A second, smaller problem

`status.code().unwrap_or(1)` at `secrets.rs:6082` turns every signal death into exit code 1. A
child killed by the OOM killer looks the same as a child that failed normally. The 137 and 143 codes
that restart policies and post mortems use are lost.

## Suggested fix

Install handlers for SIGTERM, SIGINT and SIGHUP around the `wait()` at `secrets.rs:6081` and send
the signal to the child. Return `128 + signum` when the child dies from a signal.

An `exec` mode would also solve it, because the application would replace SecretSpec and become
PID 1. That is a larger change, and it would conflict with the audit event that the current comment
describes.

## Related

Issue #211 asks how to use SecretSpec with Docker. It ends with an unverified guess about the
process model and a request to be corrected. This is the answer to that part of it.

domenkozar:
## Summary

- forward SIGTERM, SIGINT, and SIGHUP from secretspec run to its child on Unix
- return conventional 128 + signal exit codes for signal-terminated children
- document the SecretSpec 0.20+ behavior and add end-to-end regression coverage

## Testing

- devenv shell cargo test -p secretspec
- pre-commit hooks: clippy, rustfmt

Closes #382

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   [#369]: https://github.com/cachix/secretspec/issues/369
 
+- On Unix, `secretspec run` now forwards `SIGTERM`, `SIGINT`, and `SIGHUP` to
+  the command it started, allowing graceful container shutdown even when
+  SecretSpec is PID 1. Commands terminated by a signal now produce the
+  conventional `128 + signal` exit status instead of always exiting 1
+  ([#382]).
+
+  [#382]: https://github.com/cachix/secretspec/issues/382
+
 - Format-preserving `Spec` edits now keep inherited declarations separate after
   semantic builder changes, resolve parent specs independently of later working
   directory changes, remove synthesized profile tables when additions are
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5346,6 +5346,7 @@ dependencies = [
  "secrecy",
  "serde",
  "serde_json",
+ "signal-hook",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ keyring = "4.1.6"
 keepass = { version = "0.13.17", features = ["save_kdbx4"] }
 keeper-secrets-manager-core = { version = "17.3.0", default-features = false, features = ["blocking"] }
 libc = "0.2"
+signal-hook = "0.3"
 serde = { version = "1.0", features = ["derive"] }
 toml = "0.9"
 toml_edit = { version = "0.23", features = ["serde"] }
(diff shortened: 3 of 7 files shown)


moltar:
Thank you!

PR #402, opened 2026-08-26
fix(infisical): isolate Universal Auth connection

krimax409:
## Summary

- use a separate `reqwest::Client` for Infisical Universal Auth login
- keep the provider's existing pooled client for subsequent secret requests
- add a keep-alive TCP regression test that verifies login and the following secret read use different connections

The shared client can retain a connection after the Universal Auth response. Reusing that connection for the next Infisical request can leave the provider waiting while reading an otherwise successful HTTP response. Keeping the authentication client separate avoids changing pooling for normal secret reads.

## Testing

- `cargo test -p secretspec --no-default-features --features cli,infisical provider::infisical::tests -- --nocapture` (43 passed)
- `cargo build -p secretspec --no-default-features --features cli,infisical`
- `git diff --check`

The full workspace test was also attempted with one build job, but the remote runner was killed with `SIGKILL 137` during dependency compilation because of memory pressure. Existing Clippy warnings are present outside the changed files; no Clippy warning was introduced by this change. A live Infisical check of the same behavior on the earlier `v0.19.1` patch returned `39 found, 0 missing`; it is not presented as a current-main CI result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The Infisical provider now uses a separate HTTP connection for Universal Auth
+  login, keeping the connection pool used for subsequent secret reads reliable.
+
 - The value-free resolution surfaces — `secretspec check --json`,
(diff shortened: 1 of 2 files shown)


RashingPro:
👀

PR #416, opened 2026-09-01
Add Elixir SDK with precompiled NIFs

sylvesterroos:
Introduce secretspec-ex, a Rustler binding that calls the shared Rust resolver from Elixir. Add a release workflow that attaches precompiled NIF archives to GitHub Releases and publishes the package to Hex.pm, and include the SDK in workspace and devenv setup.

I've tried my best to match the layout of the other SDKs, but I might have missed something.


domenkozar (commented):
I found several release, aggregate-test, source-build, and cleanup issues. Details are inline.

.github/workflows/elixir-package.yml
```diff
@@ -0,0 +1,187 @@
...
+        run: bash scripts/sync-sdk-versions.sh
+      - name: Build Rustler NIF
+        id: build
+        uses: philss/rustler-precompiled-action@v1.1.5
+        with:
+          project-name: secretspec_native
+          project-version: ${{ env.PROJECT_VERSION }}
+          target: ${{ matrix.target }}
+          nif-version: "2.15"
+          use-cross: ${{ matrix.use-cross || false }}
```
domenkozar: **[P1] Omit the false cross input for native builds**

For the ARM Linux, macOS, and Windows jobs, `${{ matrix.use-cross || false }}` passes the non-empty string `false` to this composite action. Version 1.1.5 treats that string as truthy internally and attempts to install an x86_64-Linux `cross` binary, so these jobs fail before Cargo runs. Please omit this input for the entirely native matrix or pass an empty value.

.github/workflows/elixir-package.yml
```diff
@@ -0,0 +1,187 @@
...
+          project-dir: secretspec-ex/native/secretspec_native
```
domenkozar: **[P1] Keep Cargo output where the packaging action expects it**

Because `secretspec_native` is a member of the root workspace, Cargo writes its output under the root `target/` directory, while `rustler-precompiled-action` changes into `project-dir` and looks for `secretspec-ex/native/secretspec_native/target/<triple>/release`. The build therefore succeeds and then fails during archive creation with a missing directory. Please set a matching `CARGO_TARGET_DIR` or otherwise align the workspace output with the action.

.github/workflows/elixir-package.yml
```diff
@@ -0,0 +1,187 @@
...
+          path: artifacts
+          merge-multiple: true
+      - name: Attach NIF archives to the GitHub Release
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_TAG: ${{ inputs.release_tag || github.ref_name }}
+        run: |
+          set -euo pipefail
+          mapfile -t assets < <(find artifacts -type f -name 'libsecretspec_native-*.tar.gz' -print | sort)
```
domenkozar: **[P1] Include the Windows archive in release asset discovery**

On tag or manual release runs, the Rustler action removes the `lib` prefix for `*-pc-windows-*` and produces `secretspec_native-v...dll.tar.gz`. This glob selects only the three Unix `libsecretspec_native-*` archives, so the following count check fails and prevents both NIF publication and the dependent Hex package job.

scripts/ci-sdks.sh
```diff
@@ -191,6 +191,11 @@ run_jvm() {
   ( cd secretspec-jvm && ./gradlew build )
 }
 
+run_elixir() {
+  echo "==> Elixir"
+  ( cd secretspec-ex && SECRETSPEC_EX_BUILD=1 mix test )
```
domenkozar: **[P1] Bootstrap Mix dependencies in the aggregate SDK suite**

In the clean checkout used by the `sdks` workflow, neither Hex nor the dependencies are installed, and `mix test` exits before running tests because it cannot resolve `:jason`, `:rustler`, or `:rustler_precompiled`. The setup in the dedicated Elixir workflow is a separate job and does not carry over. Please install Hex and fetch dependencies here, and apply the same bootstrap in `conformance/run.sh`.

.github/workflows/elixir-package.yml
```diff
@@ -0,0 +1,187 @@
...
+
+  nif:
+    name: NIF ${{ matrix.target }}
+    needs: [test]
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { target: x86_64-unknown-linux-gnu, runner: ubuntu-latest }
```
domenkozar: **[P1] Build Linux archives against the documented glibc baseline**

The Linux NIFs are linked directly on Ubuntu 24.04 runners, so they can acquire a glibc 2.39 requirement and fail to load on common supported systems such as Ubuntu 22.04. This also conflicts with the `manylinux_2_28` expectation described in `docs/src/content/docs/sdk/overview.md`. Please build these archives in the baseline container/toolchain used by the other binary SDKs.

secretspec-ex/native/secretspec_native/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "secretspec_native"
+version.workspace = true
```
domenkozar, line 3: **[P2] Make the packaged native crate standalone**

When a Hex consumer forces a local Rustler build, the package contains `native/` but not the repository workspace. Cargo therefore cannot resolve these workspace-inherited fields, and the `../../../secretspec` path dependency is absent as well, so source builds fail before compilation. Please package a standalone manifest with explicit package metadata and a resolvable SecretSpec dependency.

secretspec-ex/lib/secret_spec/resolved.ex
```diff
@@ -0,0 +1,32 @@
...
+  @doc "Returns a flat map of secret names to usable values."
+  def fields(%__MODULE__{secrets: secrets}),
+    do: Map.new(secrets, fn {name, secret} -> {name, ResolvedSecret.get(secret)} end)
+
+  @doc "Removes materialized files belonging to this resolution."
+  def close(%__MODULE__{secrets: secrets}) do
+    secrets
+    |> Map.values()
+    |> Enum.filter(&(&1.as_path && is_binary(&1.path)))
+    |> Enum.each(fn secret -> File.rm(secret.path) |> ignore_missing_file() end)
```
domenkozar: **[P2] Continue cleanup after a file-removal error**

If one materialized secret cannot be deleted—for example because of permissions or a Windows sharing violation—`Enum.each` raises immediately and leaves every later secret file on disk. Cleanup should attempt all `as_path` files, retain the first error, and raise only after the remaining secret-bearing files have been removed.

.github/workflows/pre-release.yml
```diff
@@ -83,6 +85,15 @@ jobs:
       pre_release: true
       publish: false
 
+  elixir:
+    name: Elixir SDK
+    needs: authorize
+    permissions:
+      contents: read
+    uses: ./.github/workflows/elixir-package.yml
```
domenkozar, line 93: **[P2] Suppress the direct Elixir matrix in release PRs**

When a `Release ...` PR changes `secretspec-ex/mix.exs`, the Elixir workflow runs once from its own `pull_request` trigger and again through this reusable-workflow call. Unlike the other package workflows, it has no `pre_release` input or job guard, so every release PR unnecessarily runs the test plus four native builds twice.

secretspec-ex/lib/secret_spec/native.ex
```diff
@@ -0,0 +1,25 @@
...
+    otp_app: :secretspec,
+    crate: "secretspec_native",
+    targets: [
+      "x86_64-unknown-linux-gnu",
+      "aarch64-unknown-linux-gnu",
+      "aarch64-apple-darwin",
+      "x86_64-pc-windows-msvc"
+    ],
+    base_url: "https://github.com/cachix/secretspec/releases/download/v#{version}",
+    # force_build: local_checkout?,
```
domenkozar: **[P1] Enable local NIF builds before running tests**

In a source checkout there is no checksum file or published Elixir NIF for the checked-in 0.20.0 version, while RustlerPrecompiled defaults `force_build` to false. Consequently `mix test` attempts to download `libsecretspec_native-v0.20.0...` and fails with HTTP 404, blocking both the new test job and local development; enable the commented local-checkout build path while keeping published packages precompiled-only.

secretspec-ex/lib/secret_spec/builder.ex
```diff
@@ -0,0 +1,81 @@
...
+           "spec_version" => 1,
+           "base_dir" => base_dir,
+           "spec" => spec
+         },
+         "options" => request
+       }, true}
+
+  defp native_request(builder, mode) do
+    {request, versioned} = native_request(builder)
+    {Map.put(request, "mode", mode), versioned}
```
domenkozar: **[P2] Put report mode in the inline options object**

When `report/1` follows `with_inline_spec/3`, this adds `mode` at the versioned request's top level rather than under `options`. Rust's `CallRequest` denies unknown top-level fields and expects `options.mode`, so every inline report returns an `invalid_request` error instead of a report.

secretspec-ex/lib/secret_spec.ex
```diff
@@ -0,0 +1,130 @@
...
+
+      %{"ok" => false, "error" => %{"kind" => error_kind, "message" => message}} ->
+        raise %Error{kind: error_kind, message: message}
+    end
+  end
+
+  @doc false
+  defp with_default_sigchld(fun) do
+    :global.trans({__MODULE__, :sigchld}, fn ->
+      :ok = :os.set_signal(:sigchld, :default)
```
domenkozar: **[P1] Skip SIGCHLD changes on Windows**

On the advertised Windows x64 target, OTP's Windows implementation does not support `SIGCHLD`, so `:os.set_signal(:sigchld, :default)` raises `ArgumentError`. Since this executes before every NIF invocation, all resolve and report calls fail before reaching native code; restrict this workaround to Unix platforms.

secretspec-ex/lib/secret_spec.ex
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+      %{"ok" => false, "error" => %{"kind" => error_kind, "message" => message}} ->
+        raise %Error{kind: error_kind, message: message}
+    end
+  end
+
+  @doc false
+  defp with_default_sigchld(fun) do
+    :global.trans({__MODULE__, :sigchld}, fn ->
```
domenkozar: **[P1] Give each SIGCHLD lock request a unique owner**

`:global` interprets the second tuple element as the lock requester ID, so using the constant `:sigchld` allows concurrent callers to share this lock instead of blocking. With overlapping subprocess-backed resolutions, the first call to finish can restore `SIGCHLD` to ignored while the other still waits, causing its child to be reaped and `waitpid` to fail; use a requester-specific lock ID such as one containing `self()`.

CHANGELOG.md
```diff
@@ -5,15 +5,13 @@ All notable changes to this project will be documented in this file.
...
+## Unreleased
 
 ### Added
 
-- OpenPGP and OpenSSH private keys can be generated entirely in Rust (0.21+).
-  OpenPGP generation uses an explicit User ID and signing, encryption, or
-  combined capability profiles; Ed25519/Curve25519 is the default, with
-  configurable RSA available for compatibility. OpenSSH generation likewise
-  defaults to Ed25519 and supports configurable RSA keys and comments.
+- **Elixir SDK (0.21+)** (`secretspec-ex`) resolves SecretSpec declarations from
```
domenkozar, line 25: **[P2] Preserve the existing Unreleased entries**

This edit replaces the existing Unreleased/Added contents and drops the already-recorded OpenPGP/OpenSSH generation entry, leaving that Rust feature absent from the next release notes. Keep the existing content and append the Elixir entry, as required by `AGENTS.md`.

docs/astro.config.ts
```diff
@@ -466,6 +466,11 @@ Values can be resolved from: keyring (default), KeePass KDBX (0.17+), dotenv fil
               slug: "sdk/jvm",
               badge: { text: "0.20+", variant: "note" },
             },
+            {
+              label: "Elixir",
+              slug: "sdk/elixir",
+              badge: { text: "0.21+", variant: "note" },
```
domenkozar, line 426: **[P2] Add Elixir to every public SDK listing**

Adding this SDK page makes the landing page's dynamic `sdkCount` report 11, but its unchanged selector metadata and exhaustive sentence still expose only ten SDKs, while the root README also omits Elixir and the package README lacks a `0.21+` label. Add the corresponding `(0.21+)` entries and example, as required for upcoming features by `AGENTS.md`.

domenkozar:
I think this should be based on #362 otherwise we will have to maintain two protocols :(

sylvesterroos:
Sure no problem

sylvesterroos:
Alright the SDK has been updated to IPC instead of the Rust core

sylvesterroos:
The versions still have to be updated to 0.22+ right?

PR #430, opened 2026-09-04
feat(bw): support item UUID references

stegmannb:
This adds support for referencing Bitwarden Password Manager items by UUID as well as by title. UUID references are resolved exactly, avoiding ambiguity when multiple items share the same title.

## Summary

- allow Bitwarden Password Manager `ref.item` references to select an exact item UUID before full, case-insensitive title matching
- bypass `bw list items --search` for UUID reads so unrelated fuzzy hits cannot hide the addressed item

## Validation

- `devenv shell -- cargo test -p secretspec provider::bw::tests` (170 passed)
- `devenv tasks run devenv:git-hooks:run`
- `devenv shell -- npm --prefix docs run build`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/secretspec/src/provider/bw.rs
+++ b/secretspec/src/provider/bw.rs
@@ -1014,11 +1014,12 @@ fn is_note_body_field(field: &str) -> bool {
     field.eq_ignore_ascii_case("notes")
 }
 
-/// Finds the one item `item_name` addresses, for both reads and writes.
+/// Finds the one item `item_reference` addresses, for both reads and writes.
 ///
-/// Narrows the way `bw get item` itself does — by name, then by type, then
-/// refusing what is still ambiguous — with one deliberate difference: the name
-/// has to match in full.
+/// An exact item ID selects that item directly. Otherwise, narrows the way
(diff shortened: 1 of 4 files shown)


secretspec/src/provider/bw.rs
```diff
@@ -1014,11 +1014,21 @@ fn is_note_body_field(field: &str) -> bool {
     field.eq_ignore_ascii_case("notes")
 }
 
-/// Finds the one item `item_name` addresses, for both reads and writes.
+fn is_item_uuid(value: &str) -> bool {
```
domenkozar: we should use a proper uuid parsing here

stegmannb: Agreed. I initially copied the manual UUID check from the Passbolt provider

https://github.com/cachix/secretspec/blob/5ea68378f81bef710f467b329276e4e7d77dd994/secretspec/src/provider/passbolt.rs#L81-L89

I've removed the helper and now use `uuid::Uuid::parse_str(item_name).is_ok()` directly.

PR #439, opened 2026-09-10
Add Doppler provider (0.21+)

MaxDaten:
## Summary

New `doppler://PROJECT[/CONFIG]` provider behind the `doppler` feature: read, write, and delete secrets over Doppler's REST API.

The SecretSpec profile maps onto Doppler's config axis, so secret names are stored verbatim and stay readable through `doppler run`, the dashboard, and existing Doppler integrations. Pin a config in the URI (`doppler://myapp/prd`) to read every profile from one. `get_many` fetches a whole config in one request per config; Doppler's listing carries no pagination.

## Design points worth a reviewer's eye

- Values come from Doppler's `computed` field, so a secret assembled from `${...}` references arrives resolved.
- A missing secret is HTTP 200 with null fields; a missing config is a 404. The two are reported differently so an absent config cannot make every secret look unset.
- Every request names project and config explicitly. A service token repointed at another config is refused by Doppler rather than silently followed.
- Doppler's reserved `DOPPLER_PROJECT`/`DOPPLER_CONFIG`/`DOPPLER_ENVIRONMENT` names, and names or values Doppler cannot store unchanged, are refused rather than rewritten.
- A `restricted` secret Doppler withholds from a personal or CLI token is reported as a refusal, not as unset, and can still be deleted.
- Rate-limited (429) and 5xx responses are retried after Doppler's `retry-after`, up to three attempts.

## Verification

Full workspace tests, clippy, rustfmt, and both docs guards are green. The live provider suite ran green against a throwaway Doppler project. Doppler's documented behavior was confirmed for `retry-after`, `include_managed_secrets` on both listing endpoints, and `restricted` values being withheld from user-identity tokens.

### Live testing

The generic provider suite writes, so it never guesses a project. Create a throwaway Doppler project with a `default` config, then:

```bash
export DOPPLER_TOKEN=dp.st.xxx            # any token type; service or service account preferred
export DOPPLER_TEST_PROJECT=<throwaway-project>
SECRETSPEC_TEST_PROVIDERS=doppler cargo test --package secretspec provider::tests::integration_tests
```

## Known limitations for maintainers

- An unpinned `doppler://myapp` and a pinned sibling `doppler://myapp/prd` resolve to the same physical secret under profile `prd`, yet render distinct storage identities, so the cache-versus-source guard in `plan.rs` admits a cached alias that would overwrite its own source. This needs a core-side decision (pass the profile to the guard, or a provider hook); flagged here, not fixed.
- Two CI jobs sharing one `DOPPLER_TEST_PROJECT` would race in the live suite, since the Doppler project is the namespace and the harness's generated project name is unused.

## Docs

Provider page at `docs/providers/doppler`, every provider listing marked `0.21+`, changelog entry under the unreleased section.

## Contact

Happy to iterate here or on Discord: `maxdaten4k`, also in the [SecretSpec channel](https://discord.com/channels/1036369714731036712/1391835385805668432).

### DIFF
--- a/secretspec/src/provider/doppler.rs
+++ b/secretspec/src/provider/doppler.rs
@@ -523,6 +523,44 @@ fn validate_secret_value(name: &str, value: &str) -> Result<()> {
     )))
 }
 
+/// The greatest length of one `secrets=` filter, in bytes before percent
+/// encoding.
+///
+/// Doppler documents no limit on the filter itself but allows 1,200 secrets in
+/// a config, so the filter's length is bounded only by the manifest. It travels
+/// in the URI, where the bound is the API edge's request-line limit rather than
+/// anything Doppler promises, and exceeding it fails a whole profile's read at
+/// once. 4 KiB leaves ample room for the rest of the URI inside the smallest
+/// limit an HTTP front end in common use imposes.
+const MAX_FILTER_BYTES: usize = 4 * 1024;
+
+/// Splits `wanted` into comma-joined `secrets=` filters that each fit
+/// [`MAX_FILTER_BYTES`].
+///
+/// Chunking is sound because Doppler resolves `${...}` references against the
+/// config rather than against the response, so a reference whose target lands
+/// in another chunk still resolves. An empty `wanted` yields one empty filter,
+/// which reads the whole config: see [`Call::List`].
+///
+/// A single name longer than the budget still gets its own request rather than
+/// being dropped -- Doppler's own limit is 200 characters per name, so the
+/// budget cannot actually be exceeded by one.
+fn filter_chunks(wanted: &[String]) -> Vec<String> {
+    let mut chunks = Vec::new();
+    let mut chunk = String::new();
+    for name in wanted {
+        if !chunk.is_empty() && chunk.len() + 1 + name.len() > MAX_FILTER_BYTES {
+            chunks.push(std::mem::take(&mut chunk));
+        }
+        if !chunk.is_empty() {
+            chunk.push(',');
+        }
+        chunk.push_str(name);
+    }
+    chunks.push(chunk);
+    chunks
+}
+
 /// What one of Doppler's secret objects holds, classified before any policy
 /// applies.
 ///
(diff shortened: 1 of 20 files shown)


secretspec/src/provider/doppler.rs
```diff
@@ -0,0 +1,3522 @@
...
+    /// This is what makes [`same_entry`](Provider::same_entry) correct across the
+    /// two ways one secret can be addressed: `doppler://myapp` under profile
+    /// `dev` and `doppler://myapp/dev` are the same secret, and both compile to
+    /// the coordinates `dev/API_KEY`. Leaving the config in the container would
+    /// report them as different entries, and `secretspec import --delete-source`
+    /// consults exactly that before deleting -- so it would write a value to its
+    /// destination and then delete the secret it had just written.
+    ///
+    /// [`storage_identity`](Provider::storage_identity) deliberately keeps the
+    /// config: that one is compared without an address, for cache routing, where
```
domenkozar, line 1581: [P1] Reject overlapping Doppler source and cache locations

Under profile `prd`, source `doppler://myapp/prd` and cache `doppler://myapp` address the same secret but pass `plan.rs`'s storage-identity guard. This can delete authoritative data: with a personal/CLI token reading a restricted secret, `check_cache_address_is_ours` permits the unreadable address, and `cache clear` reaches Doppler's deletion path, which deliberately proceeds despite the withheld read. Reject overlapping resolved locations during cache planning; correcting `same_entry` alone does not protect this path. Please add a regression test covering these two provider URLs under profile `prd`.

MaxDaten: Fixed in planning, as suggested. `plan_one_secret` now knows the profile and, for a cached route, builds credential-free providers for the cache and each source, positions them under the profile, and refuses when the provider's own `same_entries` reports one physical entry over the addresses the cache and the source will actually use. Only a positive match refuses, like the identity guard; `ResolvedCache` carries the alias name so the refusal names what to edit.

Regression tests in `plan.rs` (feature-gated on `doppler`): `doppler://myapp/prd` with `doppler://myapp` under profile `prd` is refused in both directions; the same pairing under `dev` and two pinned sibling configs remain valid caches.

Commits: `test(plan): Doppler source and cache overlapping under a profile is refused` (red first) and `fix(plan): refuse a cache resolving to its source's entry under the profile`.

The branch is also rebased onto main after #431; the byte-native adaptation is folded into the base commit so every commit builds.

MaxDaten: Follow-up: the planning guard as first pushed did not cover the inline spelling (`uri = \"doppler://myapp/prd\"` with a `cache` table), because that route carries the alias's own name as the primary spec and the guard re-resolved specs instead of reading the resolved URI beside it. Fixed in `fix(plan): judge the inline cached alias form by its resolved primary URI`, with a red test for both pairing directions committed first.

Two related Doppler-side fixes landed alongside: `delete` now propagates a withheld probe instead of nulling the secret (the ownership check treats a read error as ours, so this is what keeps `cache clear` off a `restricted` secret at a cache address), and a batch read no longer puts reserved names into the `secrets=` filter. The remaining commits are behavior-preserving refactors of the provider.

MaxDaten:
@domenkozar thx for merging <3

PR #443, opened 2026-09-13
Agent naive implementation of Tailscale Setec provider

carterjandrew:
So, as the title says, **this was pretty much an agent run**. _I have tested this myself against a live setec server_ and the results seem promising, but that by no means indicates I understand the code, have validated security requirements, ect...

Currently I'm running a more extensive test (again using agents). But I would love to start parsing through and understanding what I've added. I'll be slowly reading the code, trying to understand what I've added and probably marking it up with questions I have along the way.

Note, I am very happy to *continue* maintaining the setec provider, this is not going to be a dump of some code that gets abandoned. I love devenv and secretspec and want to move towards becoming a contributor :)

Any help, feedback, and comments would be greatly appreciated!!


docs/src/pages/index.astro
```diff
@@ -49,6 +49,7 @@ const providerMetadata: ProviderMetadata[] = [
   { slug: 'awssm', label: 'AWS Secrets Manager' },
   { slug: 'awsps', label: 'AWS Parameter Store' },
   { slug: 'scaleway', label: 'Scaleway Secret Manager' },
+  { slug: 'setec', label: 'Tailscale Setec (0.21+)' },
```
carterjandrew, line 52: Not sure versioning would be wanted here

secretspec/src/provider/disabled.rs
```diff
@@ -72,4 +74,22 @@ mod tests {
             } if provider == "keyring"
         ));
     }
+
+    #[test]
```
carterjandrew, line 80: Not sure there would need to be tests for this here, if only keyring is being tested besides it.

docs/src/pages/index.astro
domenkozar: no versioning here

carterjandrew:
@domenkozar thanks for the changes! I'm still fairly new to rust so lot's to learn (hence this being open so long) but I'm glad to have the changes :)

secretspec/src/provider/setec.rs
```diff
@@ -0,0 +1,838 @@
...
+    async fn get_async(&self, name: &str, version: u32) -> Result<Option<SecretBytes>> {
+        let Some(response): Option<GetResponse> = self
+            .post(
+                "/api/get",
+                &GetRequest { name, version },
+                &format!("reading secret '{name}'"),
+            )
+            .await?
+        else {
+            return Ok(None);
```
carterjandrew: Not sure returning none for a 404 is very clear. Might be worth making this specific

secretspec/src/provider/setec.rs
```diff
@@ -0,0 +1,838 @@
...
+            .send()
+            .await
+            .map_err(|error| reach_error(action, error))?;
+        let status = response.status();
+        let body = response
+            .bytes()
+            .await
+            .map_err(|error| reach_error(action, error))?;
+        if status == reqwest::StatusCode::NOT_FOUND {
+            return Ok(None);
```
carterjandrew, line 254: Returning none here makes a bunch of different functions have to interpret around it, might be worth just making the error explicit at the post layer.

secretspec/src/provider/setec.rs
```diff
@@ -0,0 +1,838 @@
...
+    async fn delete_async(&self, name: &str) -> Result<bool> {
+        let info: Option<SecretInfo> = self
+            .post(
+                "/api/info",
+                &NameRequest { name },
+                &format!("checking secret '{name}' before deletion"),
+            )
+            .await?;
+        if info.is_none() {
+            return Ok(false);
```
carterjandrew, line 337: This could potentially be problematic. If someone wants to delete something and has mismatched info and delete permissions it will fail. 

I'm going to likely just remove the info call.

secretspec/src/provider/setec.rs
```diff
@@ -0,0 +1,838 @@
...
+            validate_convention_component(label, value)?;
+        }
+        let parent = format!("secretspec/{project}/{profile}");
+        Ok(match &self.config.prefix {
+            Some(prefix) => format!("{prefix}/{parent}"),
+            None => parent,
+        })
+    }
+
+    fn resolved<'a>(&self, addr: Address<'a>) -> Result<Cow<'a, NativeAddress>> {
```
carterjandrew, line 226: Just want to make sure we really need to use Cow and a lifetime here?

carterjandrew (commented):
Thanks for the input! Interesting to see, lot's to learn for me :)

PR #444, opened 2026-09-17
fix(bw): resolve a batch with one vault listing

AlexArrufat:
## Problem

The Bitwarden Password Manager provider does not override
`Provider::get_many`, so batch reads fall back to the per-request default.
Each `get` runs `bw list items`, which downloads and decrypts the whole vault,
so resolving a batch of N secrets lists the vault N times.

The 0.14.0 notes list this provider among the stores that batch through a bulk
surface, but the override was never added.

## Change

`get_many` checks authentication once, takes one listing, and resolves every
request against it, reusing `find_addressed_item` and `extract_value_from_item`
so matching and field selection stay identical to `get`. A request whose item
is absent is omitted, as in `kdbx`.

The listing is unfiltered. `get` may prefilter a name with `--search`, but a
batch can mix item IDs, which `--search` does not match, and prefiltering per
request would reintroduce the per-request invocation. It is the set `get`
already falls back to and the one `set` uses.

## Tests

`get_many_reads_the_vault_once_for_the_whole_batch` asserts a single
`list items` for three requests, two resolved values, and one absent item
omitted.

- `cargo test -p secretspec --lib provider::bw` — 187 passed
- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --exclude secretspec-php-native` — clean


domenkozar:
Pushed a follow-up commit (d5867ecc) addressing review findings on this branch:

- `get` now re-lists the vault unfiltered whenever the `--search` prefilter returns no exact match, not only when it returns nothing. A non-empty set of similarly named decoys previously made single reads (fallback chain, import verification) answer "absent" for an item that `get_many` and `set` found.
- `get_many` resolves every address before running the CLI, so a bad coordinate is reported as such rather than behind a locked-vault error or after a full listing.
- The authentication guard and unfiltered listing live in `require_authenticated` / `listed_vault`; `entry_coordinates` now uses them too. `get` and `get_many` share `lookup_in`.
- Tests: the batch test now mixes ID, case-folded name with a field, plain name, convention address and an absent item. Added batch tests for the type filter, duplicate names, bad coordinates on a locked vault, and the locked-vault error, plus a decoy prefilter test for `get`.
- Changelog entry trimmed to user-facing behavior.

This comment was created by Claude Code (https://github.com/cli/cli/issues/13904).

